PR #16, opened 2019-12-03
Removes HTS Nexus and replaces it with htsget_app

kcolford:


shaikh-rashid:
Looks good, @kcolford. Remember to add reviewers/asignees to PRs so they are notified...

shaikh-rashid (changes requested):

shaikh-rashid (approved):

PR #18, opened 2020-04-09
HPC/conda integration

shaikh-rashid:
This branch contains changes for deployment onto hpc/conda but also adds chord_drs into the deployment


lib/drs-server/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+	yaml-dev \
+	libressl-dev \
+	git \
+	postgresql-dev \
+	libffi-dev
+
+RUN mkdir /app
+WORKDIR /app
+
+RUN git clone --branch minio-backend https://github.com/c3g/chord_drs.git
```
afrosimon, line 33: I've since merged that branch into master, so you can get rid of that --branch param.

PR #36, opened 2020-05-20
Felipe add tox dep

fcoralsasso:
This PR adds tox as a way of deploying the services.


Makefile
```diff
@@ -95,6 +95,9 @@ make kubernetes
 # deploy/test all modules in $$CANDIG_MODULES using conda
 make conda
 
+# deploys all modules using Tox
+make tox
```
zbozoky, line 99: Just asking, are both conda and tox mandatory to build?

shaikh-rashid: You can use conda or tox, conda is being left in place for now...

PR #38, opened 2020-05-25
Felipe add candig server

fcoralsasso:
This PR adds Tox and Conda deployment for candig-server.


lib/candig_server/run.sh
```diff
@@ -0,0 +1,19 @@
...
+
+[ -z $CONDA_DEFAULT_ENV ] && echo 'Conda ENV not active!' && exit 1
+
+pushd $(pwd)/lib/candig_server/
+    pip install candig-server
+    pip install candig-ingest==1.3.1
+    mkdir candig-example-data
+    wget https://raw.githubusercontent.com/CanDIG/candig-ingest/master/candig/ingest/mock_data/clinical_metadata_tier1.json
+    wget https://raw.githubusercontent.com/CanDIG/candig-ingest/master/candig/ingest/mock_data/clinical_metadata_tier2.json
+    ingest candig-example-data/registry.db mock1 clinical_metadata_tier1.json
```
zbozoky, line 15: It is perfect to add some mock data to a new deployment for testing, but later this should be changed or removed, so we do not end up with a situation in which all servers holds exactly the same data.

PR #40, opened 2020-05-27
Felipe add dataset service

fcoralsasso:
This PR adds datasets service to CanDIGv2 structure.


tox.ini
```diff
@@ -84,4 +84,14 @@ deps=-r{toxinidir}/lib/rnaget/rnaget_service/requirements.txt
 
 commands=
     python setup.py develop
-    screen -dmS {envname} candig_rnaget --host={env:RNA_GET_HOST} --port={env:RNA_GET_PORT}
\ No newline at end of file
+    screen -dmS {envname} candig_rnaget --host={env:RNA_GET_HOST} --port={env:RNA_GET_PORT}
+
+[testenv:datasets]
+basepython=python3.7
```
shaikh-rashid, line 90: Just curious, why is python3.7 needed here if we are already using it for global?

tox.ini
```diff
@@ -84,4 +84,14 @@ deps=-r{toxinidir}/lib/rnaget/rnaget_service/requirements.txt
...
+changedir={toxinidir}/lib/datasets/datasets_service
+deps=-r{toxinidir}/lib/datasets/datasets_service/requirements.txt
+commands=
+    ; mkdir log
```
shaikh-rashid, line 94: maybe we should start to put this tmp files in a local tmp directory? {toxinidir}/tmp/ like this?

lib/datasets/run.sh
```diff
@@ -0,0 +1,10 @@
+set -e
+
+source .env
+
+[ -z $CONDA_DEFAULT_ENV ] && echo 'Conda ENV not active!' && exit 1
+
+pushd $(pwd)/lib/datasets/datasets_service/
+    pip install -r requirements.txt
+    python -m candig_dataset_service --host ${DATASETS_SERVICE_HOST} --port ${DATASETS_SERVICE_PORT}
```
shaikh-rashid, line 9: shouldn't this be {env: DATASETS_SERVICE_HOST} ?

tox.ini
fcoralsasso: Fixed.

tox.ini
fcoralsasso: Fixed.

PR #48, opened 2020-07-28
Monitoring services enabled

shaikh-rashid:
v0.2.4 stable version of CanDIGv2, now with monitoring and logging


afrosimon (approved):
Looking good as far as I can tell (lots of stuff in there). Thanks for the ASCII graph in the README!

PR #50, opened 2020-08-13
Stable version: v0.3.5

shaikh-rashid:
* modified chord-metadata to read external configs and secrets
* selfsigned-ssl certs are moved to tmp/ssl/
* update templates for configs and secrets, added fixes


fcoralsasso (approved):
lgtm

lib/chord-metadata/settings.py
```diff
@@ -0,0 +1,226 @@
+"""
```
afrosimon: I don't think it's a good idea in the long term to use such a file (what happens when it is modified in the repo itself?) I'd advise to set it up with a dotenv, same as chord_drs (see the candig-authz branch on chord_metadata_service, will create PR shortly).

shaikh-rashid: Agreed that there needs to be a better way. My only concern with dotenv is that though the secret is read from the `.env` file, dotenv also exposes the secret(s) as environment variable(s). This is not a good practice to have especially in the case where debugging processes echo out all environment variables to stdout/stderr. In the case like ours where the logging system is storing all debug output along with log messages, this could cause any environment variables to be saved into logs.

PR #51, opened 2020-08-20
Vault documentation

afrosimon:
So, first thing, is it the proper place for this documentation? If so, would I also include the middleware I've been building somewhere in the same repo?

### DIFF
--- a/docs/configure-vault.md
+++ b/docs/configure-vault.md
@@ -118,9 +118,10 @@ vault auth list -format=json
 # Should be in ["jwt"]["accessor"]
 ```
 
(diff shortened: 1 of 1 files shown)


ljdursi:
I'll defer to Shaikh on whether this is the right repo for the middleware, but on documentation, I'd say the first priority is to get it down and committed _somewhere_, and we can always move it elsewhere if needed.

shaikh-rashid:
I would suggest that the middleware be a standalone repo. That way it is reusable and easy to import in as a submodule.

shaikh-rashid (approved):

PR #61, opened 2020-11-20
Authz structure for the permissions JWT in Vault (COGG-103)

amanjeev:
- Adds JWT permissions structure using Vault roles via a simple (read cumbersome) template.
- Adds a document to start the conversation around the structure and its evolution.  This version using Draft v003.
- Moves the old datastructure template to new entitlements template for semantic changes

The structure of the JWT looks like 
```
{
  "aud": "cq_candig",
  "exp": 1605919533,
  "ga4gh_passport_v1": {
    "ga4gh_visa_v1": {
      "type": "ControlledAccessGrants",
      "value": {
        "dataset123": {
          "access": "4"
        },
        "dataset321": {
          "access": "4"
        }
      }
    }
  },
  "iat": 1605833133,
  "iss": "/v1/identity/oidc",
  "namespace": "root",
  "sub": "6...a0266b9"
}
```


amanjeev:
@ljdursi @brouillette @zxenia please review this PR. I think you are the best reviewers for this PR.

ljdursi (approved):
Can't really review the Vault configuration work, but the design of the permissions looks good to me.  I like that access level is a string, will let us have something more like roles rather than ordinal levels in the future.

I'd be just as happy with a list of "visas" rather than putting them into a "passport" but this works and we will have to handle passports one way or another eventually so we'll need the functionality to deal with them in OPA (and will help position us for passports 2.0)

brouillette:
LGTM

PR #62, opened 2020-11-20
Dashboard global envs

shaikh-rashid:


lib/cancogen/docker-compose.yml
```diff
@@ -2,10 +2,8 @@ version: '3.7'
 
 services:
   cancogen_dashboard:
-    tty: true
```
dashaylan, line 5: Has the cancogen container stayed running when this is removed? I was running into issues with the container exiting after starting, and attaching a shell kept it alive. 

https://dev.to/igmrrf/docker-react-exited-with-code-0-398n for reference

shaikh-rashid: You are correct, this does not work without the tty: true. This is a strange React bug...

dashaylan (approved):

fcoralsasso (approved):
Thanks.

PR #65, opened 2021-02-10
Fix pip version to 20.2.4

daisieh:
Apparently 20.3+ resolves things differently and creates a conflict between Docker and Toil.


shaikh-rashid (approved):
Great, thanks for the quick turn around on this change

PR #67, opened 2021-03-09
Update datasets: add required packages gcc libc-dev

daisieh:


shaikh-rashid (approved):

shaikh-rashid:
LGTM

daisieh:
For future searchers: these packages were required to install uWSGI on python-slim.

PR #68, opened 2021-03-10
don't build rust for cryptography

daisieh:
I don't know if we actually need the rust-built components for this particular requirement that is installed by some package we install, but turning this flag on works to allow it to build.


shaikh-rashid (changes requested):
Try installing `rust` and `cargo` via `apk` then running the build **without** disabling rust-toolchain in `pip`. We do need to compile `cryptography` with rust because the upstream developer of the library is standardizing on this build-toolchain going forward.

Ref:
https://github.com/pyca/cryptography/issues/5799
https://github.com/PyO3/pyo3/issues/1420

shaikh-rashid (approved):

PR #71, opened 2021-03-25
Updates to Vagrant, deploy OpenStack instance from Vagrant

daisieh:
All of the provisioning that was in the last version of the Vagrantfile is now broken out into separate shell scripts. provision.sh is the main provisioner, but you have to log out and log back in again to pick up the Docker group bits, so now it runs the vagrant-reload plugin and then runs setup_containers.sh, which finishes the setup.

Most of the information should be in the documentation that I've written as well. The only undocumented quirk: if you want to use the Vagrantfile but pull from a branch of CanDIGv2 that is not the main develop branch, add a second argument to line 29 or line 52 of the Vagrantfile as needed:
```
    override.vm.provision 'shell', privileged: false, path: "provision.sh", args: ["/home/vagrant/candig", "develop-shell-2"]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ services:
 
 To deploy CanDIGv2, follow one of the available install guides in `docs/`:
 
+* [Vagrant Deployment Guide (with instructions for OpenStack)](./docs/install-vagrant.md)
 * [Docker Deployment Guide](./docs/install-docker.md)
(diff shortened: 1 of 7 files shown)


shaikh-rashid (commented):
Splendid work

shaikh-rashid (approved):

PR #72, opened 2021-04-02
Track progress in vagrant provisioning; small vagrant fixes

daisieh:
I tweaked a few small things in the scripts and vagrantfile, but mostly just added tracking in a file called progress.txt and then added a lot of error checks to exit if a command fails instead of continuing.


shaikh-rashid (approved):
LGTM

PR #77, opened 2021-05-13
CanDIG release v0.5.0

shaikh-rashid:


daisieh (approved):
I'm assuming that there aren't any changes that we haven't already looked through? And it's okay that keycloak and tyk don't build quite right?

shaikh-rashid:
> I'm assuming that there aren't any changes that we haven't already looked through? And it's okay that keycloak and tyk don't build quite right?

Yes, this is more of a PR to bring stable up-to-date with develop. More changes will need to be introduced in the coming weeks/months.

PR #81, opened 2021-06-30
update cancogen-dashboard

daisieh:
I'm pretty sure this is how to update a submodule...? I needed to update the gitattributes file in that one and then pick up the changes here.


shaikh-rashid (commented):
Could you please explain why the gitattributes needed to be changed

daisieh:
Sorry! I explained it in the PR for the submodule itself, https://github.com/CanDIG/cancogen_dashboard/pull/100.

Here's the explanation:
> On machines without autocrlf set properly, the .gitattributes file could cause the line endings to get in a weird state. According to https://stackoverflow.com/questions/56857990/whats-the-difference-between-text-auto-eol-lf-and-text-eol-lf-in-gitat/56858538#56858538, fixing the line to text=auto should fix this problem.

shaikh-rashid (approved):

PR #82, opened 2021-07-08
Initial jenkins setup

daisieh:


Makefile
```diff
@@ -48,7 +48,7 @@ bin-all: bin-conda bin-docker-machine bin-kompose bin-kubectl \
 
 #<<<
 bin-conda: mkdir
-	echo "    started bin-conda" >> ~/progress.txt
+	echo "    started bin-conda" >> $(WORKING_DIR)/progress.txt
```
shaikh-rashid: There is a global $DIR variable when calling through Makefile which refers to the project directory

shaikh-rashid: Also maybe we can consider moving the progress.txt logs under $DIR/tmp/logs/session-<timestamp>.txt this way we can easily see multiple runs as logfiles

Makefile
```diff
@@ -530,7 +530,7 @@ docker-pull:
 .PHONY: docker-push
 docker-push:
 	$(foreach MODULE, $(CANDIG_MODULES), $(MAKE) push-$(MODULE);)
-	$(foreach MODULE, $(TOIL_MODULES), docker push $(DOCKER_REGISTRY)/$(MODULE):latest;)
+	#$(foreach MODULE, $(TOIL_MODULES), docker push $(DOCKER_REGISTRY)/$(MODULE):latest;)
```
shaikh-rashid: Maybe better to move this push step to the toil-docker function

etc/env/example.env
```diff
@@ -24,6 +24,9 @@ VENV_NAME=candig
 VENV_PYTHON=3.7
 VENV_PIP=20.2.4
 
+# working dir
+WORKING_DIR=~
```
shaikh-rashid: Not needed, see previous comments

provision.sh
```diff
@@ -84,8 +84,9 @@ if [ $? -ne 0 ]; then
 fi
 
 echo "  starting candig pip install" | tee -a ~/progress.txt
-source $PWD/bin/miniconda3/etc/profile.d/conda.sh
-conda activate candig
+cat "source $PWD/bin/miniconda3/etc/profile.d/conda.sh" >> ~/.profile
```
shaikh-rashid: Not recommended to do this as it can interfere with other conda environments.

setup_jenkins.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+sed 's/DOCKER_REGISTRY=.*/DOCKER_REGISTRY=ghcr.io\/candig/; s/WORKING_DIR=.*//' etc/env/example.env > .env
```
shaikh-rashid: Better to add this as instructions for user to modify in documentation for example.env.

setup_jenkins.sh
```diff
@@ -0,0 +1,16 @@
...
+  echo "need to re-run setup"
+  rm $PWD/progress.txt
+  make bin-all
+  make init-conda
+  . $PWD/bin/miniconda3/etc/profile.d/conda.sh
+  conda activate candig
+  pip install -U -r $PWD/etc/venv/requirements.txt
+else
+  echo "already set up"
+fi
```
shaikh-rashid, line 19: Don't understand why this is needed?

shaikh-rashid (changes requested):
Please see request for changes in comments

Makefile
shaikh-rashid: Also please take a look at the TODO in line 3, it would be nice to toggle this kind of debugging with a DEBUG=true flag in .env

setup_jenkins.sh
daisieh: You don't like reassurance that the other half of a conditional was reached?

Makefile
daisieh: I'm not sure that, at least the way I'm using progress.txt in my setup scripts, that this is solely a debugging log. Scripts sometimes rely on "finished xxxx" being present in order to proceed. Still, I will move it to $DIR/tmp/logs, as that is a better location for it anyway.

setup_jenkins.sh
daisieh: Are we permanently switching from Dockerhub to Github Container Registry? If so, we can just change the value directly in example.env.

shaikh-rashid (approved):
LGTM, setup_jenkins.sh makes sense for Jenkins jobs.

PR #84, opened 2021-07-13
Pin version of alpine to 3.13

daisieh:
This pins alpine to 3.13 for chord-metadata, federation-service, and chord-drs. htsget-server has its Dockerfile in its submodule, so we'll need to merge that and then update the tag here.


daisieh:
It looked like it worked when I ran it on the Jenkins setup: http://10.10.11.27:8080/job/docker-image-push/lastSuccessfulBuild/

shaikh-rashid (approved):
Nice job, thanks for updating the template as well.

PR #87, opened 2021-07-21
Switch back to Dockerhub registry

daisieh:
I haven't quite figured out how to programmatically put things into the Jenkinsfile to make it use a set of credentials/url/etc based on a registry parameter, but the rest of the backbone is there and this works for now.


daisieh:
Please ignore the 20+ commits where I tried to get the parameter thing working :/

shaikh-rashid (approved):

PR #91, opened 2021-08-18
Post Auth Merge Fixes

shaikh-rashid:
WIP Draft PR


Makefile
```diff
@@ -350,148 +380,6 @@ compose:
 	# 	| docker-compose -f - up
 
 
-#TODO: deprecate compose-authx-down
```
amanjeev, line 353: is this deprecated but still saved in `Makefile.auth.mak` ?

docs/install-docker.md
```diff
@@ -96,17 +96,16 @@ sudo usermod -aG docker $(whoami)
 
 ```bash
 # 1. initialize repo and submodules
-git clone -b stable https://github.com/CanDIG/CanDIGv2.git
+git clone -b develop https://github.com/CanDIG/CanDIGv2.git
 git submodule update --init --recursive
+cd CanDIGv2
```
amanjeev: Do we want `develop` branch for installation?

Also, I think 101 should be before 100.  I believe we have to `cd` into the cloned repo and then `submodule init`.

amanjeev:
@shaikh-rashid Is [docs/authz-setup.md](docs/authz-setup.md) still the way to deploy this setup?  It talks about `compose-authx-setup-candig-server` which is deprecated, right?

The other thing is that this seems to be deploying `candig-server` which is v1, right?  Should get rid of that in this repo and use the other services that should live behind tyk?

I am just trying to figure out how to deploy auth stuff now.  I see there are two separate directories - authorization and authentication - which is perhaps a change from using `authz` dir.

amanjeev:
I feel that I have not run the tests some of which are related to auth.  Not sure what's the best way to deal with that is.  Otherwise I am good with this.

PR #99, opened 2021-10-22
Authentication - Tyk and Keycloak

amanjeev:
This is to be merged before https://github.com/CanDIG/CanDIGv2/pull/91

- Adds Tyk and Keycloak deployment 
- Moves Tyk and Keycloak etc. to top level citizens
- Uses `Makefile.authx` to logically group authentication and authorization into a unit
- Does NOT automatically add authx to provision.sh


lib/tyk/tyk_setup.sh
```diff
@@ -0,0 +1,68 @@
...
+
+TYK_NODE_SECRET_KEY_VAL=$(cat $PWD/tmp/secrets/tyk-node-secret-key)
+export TYK_NODE_SECRET_KEY=$TYK_NODE_SECRET_KEY_VAL
+
+TYK_ANALYTIC_ADMIN_SECRET_VAL=$(cat $PWD/tmp/secrets/tyk-analytics-admin-key)
+export TYK_ANALYTIC_ADMIN_SECRET=$TYK_ANALYTIC_ADMIN_SECRET_VAL
+
+mkdir -p $CONFIG_DIR $CONFIG_DIR/apps $CONFIG_DIR/policies $CONFIG_DIR/middleware
+
+echo "Working on tyk.conf"
```
daisieh: Would you consider adding ` | tee -a $LOGFILE` to the echo commands (as in provision.sh), so that they'll end up in the progress file?

docs/authx-setup.md
```diff
@@ -0,0 +1,53 @@
...
+  TYK_EXAMPLE_API_LISTEN_PATH=/example
+  ```
+  See section `## Extra APIs can be added here`
+- Add the new section of the API to `lib/tyk/configuration_templates/policies.json.tpl` under
+  the key `access_rights`
+- Add the new section of the API tp `lib/tyk/configuration_templates/key_request.json.tpl` under 
+  the key `access_rights`
+- Add the new line to copy the file to the image in the `lib/tyk/Dockerfile`
+- Add the new line to `envsubst` in `lib/tyk/tyk_setup.sh` (see section `# Extra APIs can be added here`)
+- Redeploy the container OR use Tyk API (TODO: ask Jimmy about this)
```
daisieh, line 42: You should be able to call the POST /tyk/apis endpoint described [here](https://www.tyk.io/docs/tyk-gateway-api/)...just call it with the contents of the json file as the body and it'll hot-load it.

daisieh (changes requested):
One of my comments is more annoying than the other: basically, giant find-and-replace all of the echos with also piping the echo through `tee` to get it into the progress log file (` | tee -a $LOGFILE`).

Otherwise, as far as I can tell from my understanding of these things, it looks good.

docs/authx-setup.md
amanjeev: Thank you for this! They do not list version 2.x in the doc you provided.  I think to use this feature I first have to upgrade Tyk to 3.x and then do this.  Would you be ok if we created a jira ticket for the improvement(s) you list in this comment and tackle them later?

amanjeev: @daisieh I am going to attempt to bump tyk version and if I succeed I will report here and update this PR.  Then I will create a new ticket for improvement you suggest.  Does that sounds ok to you?

amanjeev:
I learned that I have failing health checks.  I am fixing those and will push changes soon.

amanjeev:
> I learned that I have failing health checks. I am fixing those and will push changes soon.

This is done

amanjeev:
Also, bumped up the tyk and redis version.

docs/authx-setup.md
daisieh: I see that you updated the tyk version...does that mean that we can now use the /tyk/apis endpoint? If so, can you update this doc?

amanjeev: @daisieh yes version is updated but I have no idea about the api endpoint until I try it out.  What I have tried out is file based stuff so I am not comfortable changing the docs yet.

Makefile
```diff
@@ -27,13 +28,13 @@ all:
 #<<<
 .PHONY: mkdir
 mkdir:
+	mkdir -p $(DIR)/bin
```
shaikh-rashid: repeat line

Makefile.authx
```diff
@@ -0,0 +1,55 @@
+#>>>
+# close all authentication and authorization services
+
+#<<<
+clean-authx:
+	cat $(DIR)/lib/compose/docker-compose.yml $(DIR)/lib/logging/$(DOCKER_LOG_DRIVER)/docker-compose.yml \
+		$(DIR)/lib/keycloak/docker-compose.yml | docker-compose -f - down
+
+	cat $(DIR)/lib/compose/docker-compose.yml $(DIR)/lib/logging/$(DOCKER_LOG_DRIVER)/docker-compose.yml \
+    	$(DIR)/lib/tyk/docker-compose.yml | docker-compose -f - down
```
shaikh-rashid: fix indentation

Makefile.authx
```diff
@@ -0,0 +1,55 @@
...
+
+	# - remove intermittent docker images
+	docker rmi candigv2_keycloak:latest --force
+	docker rmi candig_keycloak:latest --force
+
+	docker rmi candigv2_tyk:latest --force
+	docker rmi candig_tyk:latest --force
+
+	# - clean tmp dir inside lib/tyk
+	rm -r $(DIR)/lib/tyk/tmp
```
shaikh-rashid, line 20: does tyk tmp need to exist inside /lib/tyk ? can it be in /tmp/data/tyk ?

Makefile.authx
```diff
@@ -0,0 +1,55 @@
...
+
+
+#>>>
+# authx, common settings
```
shaikh-rashid, line 24: missing newline

etc/env/example.env
```diff
@@ -9,14 +9,6 @@ CANDIG_AUTH_MODULES=keycloak tyk opa vault arbiter
 CANDIG_DOMAIN=docker.localhost
 CANDIG_AUTH_DOMAIN=auth.docker.localhost
 
-# candig server
```
shaikh-rashid, line 12: why deprecate?

etc/env/example.env
```diff
@@ -226,31 +219,36 @@ KEYCLOAK_REALM=candig
...
 TYK_POLICY_ID=candig_policy
+## api - authentication
 TYK_AUTH_API_ID=11
 TYK_AUTH_API_NAME=Authentication
 TYK_AUTH_API_SLUG=authentication
+## api - candig-server (v1)
 TYK_CANDIG_API_ID=21
 TYK_CANDIG_API_NAME=CanDIG
 TYK_CANDIG_API_SLUG=candig
+TYK_CANDIG_API_TARGET=http://${CANDIG_DOMAIN}:${CANDIG_SERVER_PORT}
```
shaikh-rashid, line 251: you are still using candig-server env vars

etc/env/example.env
```diff
@@ -274,11 +272,6 @@ OPA_URL=http://opa:${OPA_PORT}
 
 CANDIG_AUTHZ_SERVICE_PORT=8182
 
-# same as former candig-server port - used to relay to new c-s port
```
shaikh-rashid, line 277: just want to confirm that we are deprecating arbiter code

shaikh-rashid (changes requested):
Excellent work. Minor changes and clarifications requested. Thx

Makefile.authx
shaikh-rashid: nvm i recall why this needed to be done

amanjeev: yea I just gave up early tbh

etc/env/example.env
amanjeev: not deprecated, this was a duplicate and is mentioned again below on this line https://github.com/CanDIG/CanDIGv2/blob/amanjeev/dig-633-authentication/etc/env/example.env#L193-L198

etc/env/example.env
amanjeev: yes they are define here https://github.com/CanDIG/CanDIGv2/blob/amanjeev/dig-633-authentication/etc/env/example.env#L193-L198

shaikh-rashid (approved):

etc/env/example.env
amanjeev: yes, we are removing arbiter.  

arbiter was never and is of no use as proposed.  arbiter's purpose was proposed to be the policy enforcer but that does not work unless arbiter sits in the middle of the application and the database and at the same time speaks the said database's wire protocol to munge sql request to the database.  which is not something that we are doing or will do (although this is a good research problem, right @ljdursi ? ).

amanjeev: this reminds me that i forgot to remove arbiter from the auth module list.  thanks for mentioning.

Done!

daisieh (approved):
Looks good, thanks!

PR #102, opened 2021-11-15
Update of submodules

shaikh-rashid:
Needs to be build tested


amanjeev (approved):
Given that I see the correct SHA and the changes and that Jenkins has the thing right deployments, I approve!

etc/env/example.env
```diff
@@ -118,10 +121,12 @@ MINIO_DATA_DIR=/data
 #MINIO_VOLUME_OPT+=--opt=device=/dev/sdb1
 
 # chord-drs
+#TODO: verify/update tag
```
shaikh-rashid: @daisieh

etc/env/example.env
```diff
@@ -118,10 +121,12 @@ MINIO_DATA_DIR=/data
...
 CHORD_DRS_VERSION=v0.2.0
 CHORD_DRS_PORT=6000
 
 # htsget-app
+#TODO: verify/update tag
```
shaikh-rashid: @daisieh

etc/env/example.env
```diff
@@ -180,6 +185,7 @@ FEDERATION_IP=0.0.0.0
 FEDERATION_PORT=4232
 
 # chord metadata service
+#TODO: change tag to match candig/katsu
```
shaikh-rashid: @daisieh

etc/env/example.env
```diff
@@ -250,8 +254,8 @@ TYK_CANDIG_API_NAME=CanDIG
 TYK_CANDIG_API_SLUG=candig
 TYK_CANDIG_API_TARGET=http://${CANDIG_DOMAIN}:${CANDIG_SERVER_PORT}
 
-
 # vault service
+#TODO: pin vault version
```
shaikh-rashid, line 255: @amanjeev is there a vault version we need or do we use latest?

daisieh (approved):
I updated the tags and references in example.env. It looks good to me!

PR #104, opened 2021-11-19
Tyk add new API feature + New API for Katsu/Chord Metadata

amanjeev:
Two new things added in this PR - 

1. A simple command, to redeploy Tyk after addition of new API
2. Katsu API added at the location `/katsu`

To test this, 
- deploy the CanDIGv2 stack
- run `make init-authx` 
You should see Katsu API under `http://<host:port>/katsu`

DIG-653
DIG-652


Makefile.authx
```diff
@@ -33,6 +55,7 @@ init-authx: mkdir
 	# move to a better solution so Aman made a decision to keep this hack as-is.
 	$(eval KEYCLOAK_CLIENT_ID_64=$(shell echo -n ${KEYCLOAK_CLIENT_ID} | base64))
 	@echo $(KEYCLOAK_CLIENT_ID_64) > $(DIR)/tmp/secrets/keycloak-client-local-candig-id-64
+	$(MAKE) docker-volumes
```
shaikh-rashid, line 56: You would need to run init-docker before init-authx so the docker-volumes would have already been created. Does this cause exit or are the existing volumes ignored?

Makefile.authx
```diff
@@ -1,8 +1,8 @@
 #>>>
-# close all authentication and authorization services
+# close keycloak services
```
shaikh-rashid, line 2: Please add a comment like for the make method and how to invoke it. That way it will show up in make help. See Makefile for example.

Makefile.authx
```diff
@@ -13,11 +13,33 @@ clean-authx:
 	docker rmi candigv2_keycloak:latest --force
 	docker rmi candig_keycloak:latest --force
 
+
+#>>>
+# close tyk services
```
shaikh-rashid, line 18: Add make help method in docstring

Makefile.authx
```diff
@@ -13,11 +13,33 @@ clean-authx:
 	docker rmi candigv2_keycloak:latest --force
```
shaikh-rashid: you can use `docker-compose images` to get the image id rather than doing this.

Makefile.authx
```diff
@@ -13,11 +13,33 @@ clean-authx:
...
+#>>>
+# close tyk services
+
+#<<<
+clean-tyk:
+	cat $(DIR)/lib/compose/docker-compose.yml $(DIR)/lib/logging/$(DOCKER_LOG_DRIVER)/docker-compose.yml \
+		$(DIR)/lib/tyk/docker-compose.yml | docker-compose -f - down
+
+	# - remove intermittent docker images
 	docker rmi candigv2_tyk:latest --force
```
shaikh-rashid: Use `docker-compose images` instead

Makefile.authx
```diff
@@ -13,11 +13,33 @@ clean-authx:
...
-	rm -r $(DIR)/lib/tyk/tmp
+	rm -r $(DIR)/lib/tyk/tmp || true
+
+	docker volume rm tyk-data
+	docker volume rm tyk-redis-data
+
+
+#>>>
+# close all authentication and authorization services
+
```
shaikh-rashid, line 39: Add make help method in docstring

Makefile.authx
```diff
@@ -13,11 +13,33 @@ clean-authx:
...
+#<<<
+clean-authx: clean-keycloak clean-tyk
 
 
 #>>>
```
shaikh-rashid, line 44: Add make help method in docstring

Makefile.authx
```diff
@@ -51,6 +74,43 @@ init-authx: mkdir
 	source ${PWD}/lib/keycloak/keycloak_setup.sh; \
```
shaikh-rashid, line 59: Couldn't all of this be put inside keycloak_setup.sh?

shaikh-rashid (changes requested):
Please see comments.

Makefile.authx
```diff
@@ -51,6 +74,43 @@ init-authx: mkdir
...
-	echo  ; \
-	$(MAKE) compose-tyk \
-	${PWD}/lib/tyk/tyk_key_generation.sh;
+	echo ; \
+	$(MAKE) compose-tyk; \
+	source ${PWD}/lib/tyk/tyk_key_generation.sh; \
+	echo ;
+
+#>>>
+# authx, redeploy tyk
```
shaikh-rashid, line 70: Add make help method in docstring

Makefile.authx
```diff
@@ -51,6 +74,43 @@ init-authx: mkdir
 	source ${PWD}/lib/keycloak/keycloak_setup.sh; \
 	export SERVICE=tyk; \
 	${PWD}/lib/tyk/tyk_setup.sh; \
-	echo  ; \
-	$(MAKE) compose-tyk \
-	${PWD}/lib/tyk/tyk_key_generation.sh;
+	echo ; \
+	$(MAKE) compose-tyk; \
```
shaikh-rashid, line 64: Deployment needs to support single node (compose) and multi-node (swarm/stack)

Makefile.authx
```diff
@@ -51,6 +74,43 @@ init-authx: mkdir
...
+	source ${PWD}/lib/tyk/tyk_key_generation.sh; \
+	echo ;
+
+#>>>
+# authx, redeploy tyk
+
+#<<<
+redeploy-tyk: mkdir
```
shaikh-rashid, line 74: Redeploy should be alot cleaner than this. Is there no way to simply reconfigure a running keycloak and tyk instance?

etc/env/example.env
```diff
@@ -2,7 +2,7 @@
 # - - -
 
 # site options
-CANDIG_MODULES=weavescope portainer consul traefik logging monitoring minio drs-server htsget-server chord-metadata datasets rnaget candig-server federation-service #cancogen-dashboard cnv-service wes-server jupyter igv
+CANDIG_MODULES=weavescope portainer consul traefik logging monitoring minio drs-server htsget-server chord-metadata datasets cnv-service rnaget candig-server federation-service cancogen-dashboard #wes-server jupyter igv authentication authorization
```
shaikh-rashid, line 5: I thought we are not using cnv-service as part of the default stack @ljdursi ?

amanjeev: I do not want to remove anything existing in this PR.  That is part of the reason I have DIG-654 open because there is quite a bit in the stack that needs pruning.

Makefile.authx
amanjeev: I have never seen an exit for this, the volumes are either created or skipped.  I can double check.

amanjeev: Does not matter so will skip this.  The whole authx setup assumes a basic setup is already deployed.  Is that a mistake as an assumption?

Makefile.authx
amanjeev: Keycloak is not redeployed.  

There likely is, definitely for Tyk as there are Tyk API endpoints for this.  I however found it easier to add a file and redeploy Tyk for now.

Makefile.authx
amanjeev: Can I choose to do the swarm later?  How critical is that for this _release_?

shaikh-rashid (approved):
LGTM

PR #105, opened 2021-11-19
Making changes to jenkins script(Adding htsget app)

jagdeepsason2021:


Jenkinsfile
```diff
@@ -24,6 +28,27 @@ pipeline {
                 withCredentials([usernamePassword(credentialsId: params.REGISTRY_URL, passwordVariable: 'TOKEN', usernameVariable: 'USERNAME')]) {
                     sh('. $WORKSPACE/bin/miniconda3/etc/profile.d/conda.sh; conda activate candig; echo ${TOKEN} | docker login ${REGISTRY_URL} -u ${USERNAME} --password-stdin; make docker-push OVERRIDE_REGISTRY=${REGISTRY_URL}')
                 }
+        stage('Build htsgetapp_server') {
+            steps {
+                sh '''
+                  cd ./lib/htsget-server/htsget_app
+                  ls -la
```
amanjeev, line 35: Minor: Is this listing a requirement?

amanjeev (approved):
LGTM but I would prefer others to review before this is merged.

Jenkinsfile
```diff
@@ -24,6 +28,27 @@ pipeline {
...
+                  bash htsget.sh
+                   '''
+                }
+            }
+        }
+      post {
+  //  Status message gets sent to the Slack Channel after the job finishes
+        always {
+          success {
+            slackSend channel: 'automation-email', message:"Build deployed successfully" -${currentBuild.currentResult} ${env.JOB_NAME} ${env.BUILD_NUMBER} ${env.BUILD_URL}
```
shaikh-rashid, line 45: I think the failure notification is useful, but receiving daily emails (based on the cron job cycle) for successful build may be too noisy.

shaikh-rashid (changes requested):
Please review comments @jagdeepsason2021

Jenkinsfile
jagdeepsason2021: yes Aman. I created a bash script and wanted the script to go in, check the folder and then run and execute jenkins pipeline script.

Jenkinsfile
jagdeepsason2021: Yes @shaikh-rashid that make sense but it is only sending to the automation channel. I think it will be meaningful so we can click on the link provided and see the results in the output.

shaikh-rashid (approved):

shaikh-rashid:
@jagdeepsason2021 what is the status of this PR?

PR #109, opened 2022-01-05
Add shared volume to docker stack

daisieh:
We can remove this if necessary for production, but it seems very useful for development.


lib/chord-metadata/docker-compose.yml
```diff
@@ -57,6 +57,8 @@ services:
       #interval: 30s
       #timeout: 20s
       #retries: 3
+    volumes:
+      - shared-data:/shared
```
shaikh-rashid, line 61: This will work but the issue is that using a shared-data volume across multiple containers may result in file system i/o issue. Especially if you have different containers using the same file path with the shared volume

lib/compose/docker-compose.yml
```diff
@@ -38,6 +38,7 @@ volumes:
     external: true
   vault-data:
     external: true
+  shared-data:
```
shaikh-rashid, line 41: This should be an external volume

shaikh-rashid (changes requested):
Not sure what the purpose of using the shared volume across all these containers helps to resolve. Maybe we need to have a discussion offline to clarify this.

lib/chord-metadata/docker-compose.yml
daisieh: I think this is a temporary state: for now, it's possible that this will cause a conflict between htsget and candig-server, but as I switch this to minio, that shouldn't be a problem. Otherwise, this is primarily a staging ground for ingest data and won't be used again after the ingest.

daisieh:
After chatting with Shaikh, I decided that this isn't necessary after all. The ingest now copies the files directly to each container.

PR #114, opened 2022-02-07
update make target

daisieh:


amanjeev:
Thank you for catching that!

amanjeev (approved):

PR #119, opened 2022-03-08
Integrating OPA into the stack

daisieh:
Set up the stack as described in docs/install-docker.md.

To test:

* Get a user token, where the values for the data parameters are found in the files in `tmp/secrets`:
```
curl -X "POST" "http://auth.docker.localhost:8080/auth/realms/candig/protocol/openid-connect/token" \
     -H 'Content-Type: application/x-www-form-urlencoded; charset=utf-8' \
     --data-urlencode "client_id=local_candig" \
     --data-urlencode "client_secret=<value in keycloak-client-local_candig-secret>" \
     --data-urlencode "grant_type=password" \
     --data-urlencode "username=<value in keycloak-test-user>" \
     --data-urlencode "password=<keycloak-test-user-password>" \
     --data-urlencode "scope=openid"
```
* Using the access_token from that, get the results from:
```
curl "http://docker.localhost:5080/katsu/api/phenopackets" \
     -H 'Authorization: Bearer <access_token>'
```
You should have 4 datasets, including registered3 and controlled4.

If you follow the same steps but with the user/pass for keycloak-test-user-2, you should get 3 datasets, including controlled5 (but not registered3 or controlled4).


Makefile
```diff
@@ -395,7 +395,7 @@ compose-%:
 	echo "    started compose-$*" >> $(LOGFILE)
 	cat $(DIR)/lib/compose/docker-compose.yml $(DIR)/lib/logging/$(DOCKER_LOG_DRIVER)/docker-compose.yml \
 		$(DIR)/lib/$*/docker-compose.yml \
-		| docker-compose -f - up -d $(SERVICE)
```
shaikh-rashid, line 398: The service variable is needed in the case that your docker-compose has multiple containers/services.

shaikh-rashid (approved):
Great work, looks ready to merge

Makefile
daisieh: Yeah, but it causes problems if what you want is to spin up all of the containers in docker-compose (as is the case for opa)

PR #127, opened 2022-04-06
Building with Vagrant on VirtualBox no longer works

sdumitriu:
The base image in use was too old (debian buster), and docker-compose was not installed


shaikh-rashid (approved):

daisieh (approved):
LGTM

PR #128, opened 2022-04-08
Miscellaneous changes to server deployment settings

daisieh:
A bunch of tweaks to settings and whatnot, based on what we worked on in candig-dev deployment.

PR #129, opened 2022-04-08
The candig data portal doesn't connect to the right endpoint by default

sdumitriu:
The `ARG`s specified before `FROM` are only usable in the `FROM` line, so they should be defined after the `FROM` line.

The `ENV` variables were not the the ones that the data portal is using.


shaikh-rashid:
@daisieh I believe these have been addressed in https://github.com/CanDIG/CanDIGv2/pull/128

daisieh:
Yes, this shouldn't be needed anymore. Closing.

PR #132, opened 2022-04-14
Building vault fails because the configuration is not built yet

sdumitriu:


daisieh:
I don't think this is necessary...the compose will build the container if it's not yet built.

sdumitriu:
It does, but the build fails because `vault-config.json` is not yet generated. That's why I moved the build step after the `envsubst` generates the config.

sdumitriu:
Or do you mean that the build step can be removed completely?

PR #131, opened 2022-04-14
More tweaks

daisieh:

PR #134, opened 2022-04-14
Make https certificates if not already present since they are required

sdumitriu:


shaikh-rashid (approved):

daisieh:
This would be better as a change to Makefile. I've added it to https://github.com/CanDIG/CanDIGv2/pull/131/commits/6c3d3aea82a7c932c24e6054971ea0e34c33be4f

PR #135, opened 2022-04-14
Fix opa container name in the opa_setup script

sdumitriu:


shaikh-rashid (commented):
This will need to be more generic as the namespace (ie. candigv2 or candig) is dependant on the repo name when cloned.

shaikh-rashid (approved):

daisieh:
I've made this more generic in https://github.com/CanDIG/CanDIGv2/pull/131/commits/6540b399e1c37535d1ce5c9eb9348804e60f7288

PR #136, opened 2022-04-14
Build the auth containers as well on VirtualBox

sdumitriu:


sdumitriu:
Not sure if the `setup_auth` script should be executed in any of the other providers (libvirt/openstack).

shaikh-rashid (approved):

PR #137, opened 2022-04-20
DIG-663 & DIG-763 - Adding Vault Helper Tool to Candigv2 & Documentation for Running VHT

Zamm178:
## Description
This PR deals with the ticket [DIG-663](candig.atlassian.net/browse/DIG-663) and [DIG-763](candig.atlassian.net/browse/DIG-763). The Vault Helper Tool is now added as a git submodule to the `lib/vault` subdirectory. 
Note that the current implementation does not use the bearer token as  the vault api package used here does not allow for authorization with a bearer token
I also attempted to add the vault helper tool as a docker container since not everyone has golang installed, but this was left out for timing purposes. So, golang should be installed prior to running the tool.
The corresponding changes to the Vault Helper Tool can be seen in the [corresponding PR](https://github.com/CanDIG/Vault-Helper-Tool/pull/8).
## Changelog
- added VHT as git submodule 
- added smoke testing guide to repo in `docs/Test-VHT.md`
## Dependencies
Needs to be run from the root of the `Candigv2` repo.
## Testing
First add the token to the token.txt file.
Then either run 
```
go install -v https://github.com/CanDIG/Vault-Helper-Tool
```
or run the following commands from the root of the Candigv2 repo to copy over the binary file. (This is necessary since the paths in the settings are configured to work for Candigv2). 
```
cd lib/vault/Vault-Helper-Tool/cli \
go build \
cd - \
cp lib/vault/Vault-Helper-Tool/cli/cli .
```
Then, from the root of the `Candigv2` repo, run 
```
./cli <command> <optional-args>
```


.gitmodules
```diff
@@ -34,3 +34,6 @@
 [submodule "lib/graphql/GraphQL-interface"]
   path = lib/graphql/GraphQL-interface
   url = https://github.com/CanDIG/GraphQL-interface.git
+[submodule "lib/vault/Vault-Helper-Tool"]
```
shaikh-rashid: Installing via `go install` is sufficient. In this case I would suggest not including the helper tool as a submodule.

daisieh (approved):
I can't tell from this PR why it still thinks the submodule is part of the PR, but I think I'll have to merge it to see if it's really there.

PR #140, opened 2022-05-30
Update minio container

daisieh:
Minio container should be launched in the stack now and usable via htsget.

You should be able to view the minio console at docker.localhost:9090 (with the minio-access-key and minio-secret-key as the username/password). If you run pytest on the htsget container (you may have to install pytest: `pip install pytest`), you should find that it makes the test objects in the minio bucket.


Makefile
```diff
@@ -669,11 +669,21 @@ ssl-cert:
 	openssl req -new -key $(DIR)/tmp/ssl/selfsigned-site.key \
 		-out $(DIR)/tmp/ssl/selfsigned-site.csr -sha256 \
 		-subj '/C=CA/ST=ON/L=Toronto/O=CanDIG/CN=CanDIG Self-Signed Cert'
+	sed -i .tmp s/CANDIG_DOMAIN/$(CANDIG_DOMAIN)/ $(DIR)/etc/ssl/site.cnf
```
shaikh-rashid: Instead of `sed` can this CANDIG_DOMAIN be set with the same value as `.env`?

Makefile
```diff
@@ -609,7 +609,7 @@ minikube:
 # make minio-secrets
 
 #<<<
-minio-secrets:
+minio-secrets: ssl-cert
```
shaikh-rashid: since `ssl-cert` is already run with `init-docker`, do we need to run it again here?

Makefile
shaikh-rashid: Also where does `.tmp` come from?

lib/htsget-server/docker-compose.yml
```diff
@@ -10,6 +10,8 @@ services:
         opa_secret: "${CANDIG_OPA_SECRET}"
         opa_url: "${OPA_URL}"
         candig_auth: "${CANDIG_AUTHORIZATION}"
+        minio_url: "${CANDIG_DOMAIN}:${MINIO_PORT}"
```
shaikh-rashid: MinIO port will not be accessible in prod. Accessing MinIO will need to go through Tyk

lib/htsget-server/docker-compose.yml
```diff
@@ -29,5 +31,14 @@ services:
...
+    secrets:
+      - source: minio-access-key
+        target: access
+      - source: minio-secret-key
+        target: secret
+      - source: selfsigned-site-pem
+        target: cert.pem
     logging: *default-logging
+    environment:
+      HTSGET_TEST_KEY: "hoodlebug"
```
shaikh-rashid, line 43: What is TEST_KEY for?

etc/ssl/alt_names.txt
```diff
@@ -0,0 +1 @@
+subjectAltName = DNS:docker.localhost
```
shaikh-rashid: Need clarification on this

lib/htsget-server/docker-compose.yml
```diff
@@ -29,5 +31,14 @@ services:
         - "traefik.docker.lbswarm=true"
         - "traefik.http.routers.htsget-app.rule=Host(`htsget-app.${CANDIG_DOMAIN}`)"
         - "traefik.http.services.htsget-app.loadbalancer.server.port=${HTSGET_APP_PORT}"
+    secrets:
+      - source: minio-access-key
+        target: access
+      - source: minio-secret-key
+        target: secret
+      - source: selfsigned-site-pem
```
shaikh-rashid, line 39: Why is ssl cert needed for htsget?

lib/minio/minio/setup.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
shaikh-rashid, line 1: Will need clarification on use of ssl cert for minio and htsget

shaikh-rashid (changes requested):
Need to follow up with @daisieh regarding PR

Makefile
daisieh: (the .tmp was from a weird issue that I was having with my local version of sed that doesn't seem to happen on dev? So I am taking it back out)

Isn't the sed interpolating the value from .env? I'm not sure what you're suggesting to do instead...?

Makefile
daisieh: probably not...I'll take that out

lib/htsget-server/docker-compose.yml
daisieh: sorry, this is clearer if you're on the correct version of htsget. For htsget's pytest, you need to have a test key to bypass OPA, and it's set as an env var that both the htsget server and its pytest have access to.

daisieh: (I updated the commit for htsget below)

etc/ssl/alt_names.txt
daisieh: Sorry, this was also supposed to be CANDIG_DOMAIN

lib/htsget-server/docker-compose.yml
daisieh: If htsget is to access our self-signed minio instance (for dev setups), it needs the cert.

daisieh: It gets picked up by the correct version of htsget in its entrypoint.sh.

lib/htsget-server/docker-compose.yml
daisieh: This will get deprecated pretty quickly in https://github.com/CanDIG/CanDIGv2/pull/142: after I move S3 stuff (including local Minio) to be accessed through Vault, this isn't needed anymore.

PR #143, opened 2022-06-07
Changes required in AuthX stack, bug fixes and tweaks 🐛🚀

shaikh-rashid:


daisieh (approved):
I deployed these changes on dev and it looks good. I had to tweak one file (opa/docker-compose) because the env var was `$KEYCLOAK_URL` instead of `$KEYCLOAK_PUBLIC_URL`, but I think that was probably a typo.

PR #149, opened 2022-06-21
pin python version and alpine version for katsu📌

shaikh-rashid:


shaikh-rashid:
Minor change does not require review

PR #151, opened 2022-06-27
Small fixes

daisieh:
The only change that was breaking on prod before was the OPA_PRIVATE_URL one, but I also added in env vars for DB_PATH as an external docker volume and CANDIG_DEBUG_MODE.


etc/env/example.env
```diff
@@ -9,6 +9,7 @@ CANDIG_AUTH_MODULES=keycloak tyk opa vault
 CANDIG_DOMAIN=docker.localhost
 CANDIG_AUTH_DOMAIN=docker.localhost
 CANDIG_SITE_LOCATION=UHN
+CANDIG_DEBUG_MODE=1
```
shaikh-rashid, line 12: Is debug what we want by default?

shaikh-rashid (changes requested):
Please merge htsget_app in first, checkout new tag, and update htsget version in example.env

etc/env/example.env
daisieh: I think so...I'd imagine that prod is the only one that we wouldn't want on debug mode.

shaikh-rashid (approved):

PR #154, opened 2022-08-02
Add switches to Makefile to download M1 binaries

daisieh:
This is probably not quite correct in the case of someone trying to install onto an M1 machine with Linux, but for MacOS, this should work.

I also cleaned up the bin-all target and removed deprecated kubernetes stuff.

Don't forget to change the value of VENV_OS in your .env file to `arm64mac`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ endif
 bin-traefik: mkdir
 	echo "    started bin-traefik" >> $(LOGFILE)
 ifeq ($(VENV_OS), arm64mac)
-	curl -Lo $(DIR)/bin/miniconda_install.sh \
+	curl -Lo $(DIR)/bin/traefik.tar.gz \
 		https://github.com/traefik/traefik/releases/download/v$(TRAEFIK_VERSION)/traefik_v$(TRAEFIK_VERSION)_darwin_arm64.tar.gz
 else
 	curl -Lo $(DIR)/bin/traefik.tar.gz \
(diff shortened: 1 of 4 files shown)


SonQBChau:
Worked! be able to run the makefiles on new mac m1 :tada:

Makefile
```diff
@@ -159,8 +135,13 @@ bin-prometheus: mkdir
 #<<<
 bin-traefik: mkdir
 	echo "    started bin-traefik" >> $(LOGFILE)
+ifeq ($(VENV_OS), arm64mac)
+	curl -Lo $(DIR)/bin/miniconda_install.sh \
```
kcranston: the destination path should be /bin/traefik.tar.gz not /bin/miniconda_install.sh

Makefile
```diff
@@ -159,8 +135,13 @@ bin-prometheus: mkdir
...
+		https://github.com/traefik/traefik/releases/download/v$(TRAEFIK_VERSION)/traefik_v$(TRAEFIK_VERSION)_darwin_arm64.tar.gz
```
kcranston, line 140: and the filename contains `darwin_amd` not `darwin_arm`

daisieh: No, this one is supposed to be darwin_arm; that's the M1 mac.

Makefile
daisieh: thank you for catching that!

Makefile
kcranston: Doh. There is no arm version for 2.5.0, so we need to update to https://github.com/traefik/traefik/releases/download/v2.5.1/traefik_v2.5.1_darwin_arm64.tar.gz

kcranston: Right. The darwin_arm version doesn't appear until traefik 2.5.1, so need to add a note to docs that M1 users need to update TRAEFIK_VERSION in their .env to at least 2.5.1 (alternately, we could update the traefik version in the example .env file).

kcranston:
I would definitely like a review from @shaikh-rashid on the `docs/install-docker.md` changes

shaikh-rashid (approved):

PR #155, opened 2022-08-02
Update python and pip version to Apple Silicon

SonQBChau:
This PR based on #154 

- Python bump from 3.7 to 3.9
- Pip bump from 20.2.4 to 21.2.2
This should resolve the error packages not available in conda channels.


daisieh (approved):
I think this looks fine, but let's hold off on merging until after Shaikh reviews #154, since this subsumes that one.

shaikh-rashid (approved):

PR #158, opened 2022-09-08
Update submodules

daisieh:
Catching up with changes


shaikh-rashid:
If these are new versions of submodules, could you please also update the versions of the services in etc/env/example.env. The submodules should also be based on tagged releases/branches of their respective repos.

shaikh-rashid (changes requested):
Update example.env and include version tags for repos

PR #160, opened 2022-09-23
fix: add compatibility

SonQBChau:
For Docker Desktop 1.x use _ but 2.x use - when naming. This will option retain the compose compatibility


daisieh (approved):
LGTM

PR #162, opened 2022-09-30
pass container name into chord_metadata

daisieh:
We need to pass in the environment variable HOST_CONTAINER_NAME into the chord-metadata container.

Copy over the change in example.env to your .env.


SonQBChau (approved):
working for me

PR #170, opened 2022-10-20
Sonchau/install docker m1

SonQBChau:
update instruction for mac m1, now using docker.localhost for candig setting


daisieh (approved):
LGTM, assuming that I resolved the merge conflicts correctly.

PR #172, opened 2022-10-24
Downgrade python to 3.9

shaikh-rashid:


daisieh (commented):
I'm looking to see if downgrading will cause problems with katsu etc.

daisieh (commented):
We should update federation-service instead of universally downgrading python, since UHN's security audit specifically calls out pythons less than 3.10. I fixed the build error for federation-service in https://github.com/CanDIG/federation_service/pull/28.

daisieh:
We should fix federation-service instead of downgrading to a known insecure python: https://github.com/CanDIG/federation_service/pull/28

PR #176, opened 2022-11-22
restored cancogen_dashboard envs for use

shaikh-rashid:


daisieh:
Let's leave this unmerged for now; I don't know that we actually want to re-integrate cancogen dashboard, but it doesn't hurt to have an easy-to-find branch for it.

shaikh-rashid:
Agreed, PR will be closed but branch will remain open for future use

PR #180, opened 2022-11-30
DIG-931: Vault aws policy needs update permissions

daisieh:
Also added in an env var to be passed in for htsget; it is optional so no change to test.

### DIFF
--- a/lib/htsget-server/docker-compose.yml
+++ b/lib/htsget-server/docker-compose.yml
@@ -44,6 +44,7 @@ services:
       HTSGET_TEST_KEY: "hoodlebug"
       HTSGET_URL: ${TYK_LOGIN_TARGET_URL}/${TYK_HTSGET_API_LISTEN_PATH}
       OPA_URL: ${OPA_PRIVATE_URL}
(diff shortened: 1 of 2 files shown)


lib/htsget-server/docker-compose.yml
```diff
@@ -44,6 +44,7 @@ services:
       HTSGET_TEST_KEY: "hoodlebug"
       HTSGET_URL: ${TYK_LOGIN_TARGET_URL}/${TYK_HTSGET_API_LISTEN_PATH}
       OPA_URL: ${OPA_PRIVATE_URL}
+      OPA_SITE_ADMIN_KEY: ${OPA_SITE_ADMIN_KEY} # only required if not equal to "site_admin"
```
shaikh-rashid, line 47: If this is a secret, it should be provided via `secret` and not `environment`

shaikh-rashid (changes requested):
Request for clarification

lib/htsget-server/docker-compose.yml
daisieh: It's not a secret, it's just the value "site_admin" usually, but I wanted to leave it open for a site to have a different role name if they choose.

shaikh-rashid (approved):
Reviewed and approved

PR #181, opened 2022-11-30
Remove candig server and update module list

kcranston:
Removing candig-server from the example .env file and updating the readme with the minimal and production list of modules.


shaikh-rashid (approved):
LGTM

PR #183, opened 2022-12-02
Documentation for WSL federation configuration

CourtneyGosselin:
## Description
Documentation pertaining to the WSL federation configuration regarding WSL Docker and Windows Docker communicating with the tmp folder.


kcranston:
It looks like that federation error in WSL is repeated in two files? I'd prefer keeping it in one place, and then one file can refer to the other (otherwise, we need to maintain both places in case of changes).

CourtneyGosselin:
@kcranston moved it to just one file and referenced the file in the other document

kcranston (approved):

PR #184, opened 2022-12-07
Module cleanup

shaikh-rashid:


.gitmodules
```diff
@@ -7,27 +7,12 @@
...
-[submodule "lib/cancogen-dashboard/cancogen_dashboard"]
-  path = lib/cancogen-dashboard/cancogen_dashboard
-  url = https://github.com/CanDIG/cancogen_dashboard.git
 [submodule "lib/candig-data-portal/candig-data-portal"]
   path = lib/candig-data-portal/candig-data-portal
   url = https://github.com/CanDIG/candig-data-portal.git
 [submodule "lib/opa/opa"]
   path = lib/opa/opa
   url = https://github.com/CanDIG/candig-opa.git
-[submodule "lib/graphql/GraphQL-interface"]
```
daisieh, line 31: I think we might be bringing this one back soon...?

Makefile
```diff
@@ -518,78 +311,7 @@ init-conda:
...
-# make minikube
-
-#<<<
-.PHONY: minikube
-minikube:
-	$(DIR)/bin/minikube start --container-runtime $(MINIKUBE_CRI) \
-		--cpus $(MINIKUBE_CPUS) --memory $(MINIKUBE_MEM) --disk-size $(MINIKUBE_DISK) \
-		--network-plugin cni --cni $(MINIKUBE_CNI) --driver $(MINIKUBE_DRIVER) \
-		--dns-domain $(CANDIG_DOMAIN) --nodes $(MINIKUBE_NODES)
+init-docker: docker-volumes docker-secrets
```
daisieh, line 309: Why are we taking out ssl-cert?

Makefile
```diff
@@ -603,8 +325,6 @@ minio-secrets:
 	@echo '[default]' > $(DIR)/tmp/secrets/aws-credentials
 	@echo "aws_access_key_id=`cat tmp/secrets/minio-access-key`" >> $(DIR)/tmp/secrets/aws-credentials
 	@echo "aws_secret_access_key=`cat tmp/secrets/minio-secret-key`" >> $(DIR)/tmp/secrets/aws-credentials
-	cp $(DIR)/tmp/ssl/selfsigned-site.crt $(DIR)/tmp/secrets/selfsigned-site-crt
```
daisieh, line 606: Have you checked that local minio works without ssl certs?

daisieh (changes requested):
Overall, I think this is pretty good, but there's a few weird things that need clarification.
* I think we're bringing back rnaget and graphql eventually; should we leave those in?
* I can't tell from this PR alone why logging was removed?
* Minio was using some of the ssl-certs things; have you tested to see if it works still without them?

.gitmodules
shaikh-rashid: The graphql module was never in the list of modules in example.env and also rnaget is not a fully usable module (it might also make sense to bundle it with htsget). Both repos will require refactoring and can be reintroduced when needed in the future.

Makefile
shaikh-rashid: We no longer need self-signed certs. In the cases where we do use SSL, it is handled by an external ingress/proxy.

Makefile
shaikh-rashid: Yes, by changing console-address to http://${CANDIG_DOMAIN:{MINIO_UI_PORT} the console can be reached without ssl

daisieh: It works without ssl even with the port set as 9090 as before, even without the MINO_CONSOLE_ADDRESS env var in the docker-compose. I think it was glitching before because of the compose-% target change.

daisieh:
I added a few tweaks to revert the many docker-compose calls back to more like the way they were, minus the logging, because otherwise it is inconsistent in trying to name the stack "compose" because that's the first location for the first docker-compose file. Also for the main compose-% target, we need to keep the --compatibility flag to keep the container names from switching from underscores to hyphens.

I also did a few tweaks to opa_setup and vault_setup because they weren't quite waiting for things to launch right.

daisieh (changes requested):
I think a bunch of stuff is better now, but backing out with make clean-all and make clean-authx don't work quite right without the network specifications for them being external: I think that without that specification in lib/compose/docker-compose.yml and the docker-networks target in make, all of the clean commands try to delete the network and then error out because they can't, because it should be external.

Edited to note that because make clean-all runs all of the docker-compose downs as part of a single target, the error doesn't kill the whole process, but make clean-authx calls subtargets, and when those error out, the whole process dies.

Makefile.authx
```diff
@@ -33,8 +31,7 @@ clean-vault:
 
 #<<<
 clean-opa:
-	cat $(DIR)/lib/compose/docker-compose.yml $(DIR)/lib/logging/$(DOCKER_LOG_DRIVER)/docker-compose.yml \
-		$(DIR)/lib/opa/docker-compose.yml | docker-compose -f - down
+	docker-compose -f $(DIR)/lib/compose/docker-compose.yml docker-compose -f $(DIR)/lib/opa/docker-compose.yml down || true
```
daisieh: need to remove the second docker-compose command: should be `docker-compose -f $(DIR)/lib/compose/docker-compose.yml -f $(DIR)/lib/opa/docker-compose.yml down || true`

daisieh (changes requested):
One typo left, I think. Overall it works now: I did `make compose` and `make init-authx`, checked to see that things were connected, then did `make clean-authx` and `make clean` and it removed everything as expected (once I fixed the one typo).

I'm still not thrilled that the docker stack is now named `compose` because of that first lib/compose/docker-compose.yml, but what if we renamed that directory to lib/*candigv2*/docker-compose.yml? Then the stack would always be named candigv2, even on prod, where the repo isn't checked out in a directory of that name.

Makefile.authx
shaikh-rashid: Done

daisieh (approved):
Looks good! Everything seems to work: am able to get bearer tokens and pass them via Tyk to get datasets from Opa.

PR #188, opened 2022-12-09
DIG-996: Automate /etc/hosts step during init-authx

OrdiNeu:
[Jira link](https://candig.atlassian.net/browse/DIG-996)

This automates the inclusion of the `/etc/hosts` file into the `make init-authx` step. Includes the removal of documentation related to the step.

To test:
- [x] Linux build
- [x] Windows build
- [x] MacOS build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,6 +244,13 @@ docker-secrets: mkdir minio-secrets
 	$(MAKE) secret-opa-root-token
 	$(MAKE) secret-opa-service-token
 
+#>>>
+# modify the hosts file
+
+#<<<
+.PHONY: hosts-file
+hosts-file:
+	source ${PWD}/setup_hosts.sh
 
 #>>>
 # create persistant volumes for docker containers
@@ -306,7 +313,7 @@ init-conda:
 
 #<<<
 .PHONY: init-docker
-init-docker: docker-volumes docker-secrets
+init-docker: docker-volumes docker-secrets hosts-file
 
 
 #>>>
--- a/Makefile
+++ b/Makefile
@@ -248,8 +248,8 @@ docker-secrets: mkdir minio-secrets
 # modify the hosts file
 
 #<<<
-.PHONY: hosts-file
-hosts-file:
+.PHONY: init-hosts-file
+init-hosts-file:
 	source ${PWD}/setup_hosts.sh
 
 #>>>
@@ -313,7 +313,7 @@ init-conda:
 
 #<<<
 .PHONY: init-docker
-init-docker: docker-volumes docker-secrets hosts-file
+init-docker: docker-volumes docker-secrets
 
 
 #>>>
--- a/docs/install-docker.md
+++ b/docs/install-docker.md
@@ -1,7 +1,7 @@
 # CanDIGv2 Install Guide
 
 ---
-These instructions work for server deployments or local linux deployments. For local OSX using M1 architecture, follow the [Mac Apple Silicon Installation](#mac-apple-silicon-installation) instructions at the bottom of this file. For WSL you can follow the linux instructions and follow WSL instructions for hosts file at [update hosts](#update-hosts).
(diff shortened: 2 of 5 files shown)


docs/install-docker.md
```diff
@@ -242,28 +294,7 @@ make init-docker
 make compose
 ```
 
-### Step 5: Update hosts
```
shaikh-rashid, line 245: I am not sure what is going on but your doc is out of date. This is adding changes that were removed from previous PR

docs/install-docker.md
```diff
@@ -136,13 +128,73 @@ make docker-pull
...
 make compose
-make init-authx # If this command fails, try the #update-hosts section of this Markdown file
+make init-authx # If this command fails, try the #update-firewall section of this Markdown file
 # TODO: post deploy auth configuration
 
 ```
 
 If the command still fails, it may be necessary to disable your local firewall, or edit it to allow requests from all ports used in the Docker stack.
 
+## Option 2: Deploy CanDIGv2 using Docker Swarm
```
shaikh-rashid: This section was removed from previous PR.

shaikh-rashid (changes requested):
Please review git log and review changes. It may be that your doc is adding sections back that was removed from previous PR

daisieh (changes requested):
Even though vault_setup is currently the place where this issue first manifests, it's actually just coincidence that no other container tries to call another container using the docker.localhost alias in an earlier step. It's not specifically vault-related.

This code change would probably be best made as a new target in Makefile that gets called as, perhaps, part of the init-docker target.

setup_hosts.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Early-abort if this step isn't required
+test=$(grep "	docker\.localhost" /etc/hosts)
```
daisieh: Take out the tab in the grep phrase: it's not always present.

setup_hosts.sh
```diff
@@ -0,0 +1,35 @@
...
+
+numlines=$(echo "$IP" | wc -l)
+if [ "$numlines" == "1" ]; then
+    printf "$IP\tdocker.localhost" >> .hosts.tmp
+    sudo mv .hosts.tmp /etc/hosts;
+else
+    echo "More than one IP has been detected. Since this script can't automatically determine which one"
+    echo "will make the build work, please remove all but one of the inserted docker.localhost lines in /etc/hosts."
+    echo $IP >.hosts.tmp2
+    awk '{printf "%s\tdocker.localhost\n", $0}' .hosts.tmp2 >> .hosts.tmp
```
daisieh, line 32: I think you're going to need to loop this somehow; otherwise you just get $IP followed by the docker.localhost part.

daisieh (changes requested):

setup_hosts.sh
OrdiNeu: As far as I can tell, `awk` should include it after every line. What version of `awk` do you have installed?

daisieh: awk version 20200816

daisieh (approved):
Looks good! Works perfectly on my system.

CourtneyGosselin:
On windows though we also need to change the windows hosts file when using WSL. Maybe we could add a line that tells the user they need to add the IP added to WSL hosts to the windows as well.

Makefile
```diff
@@ -306,7 +313,7 @@ init-conda:
 
 #<<<
 .PHONY: init-docker
-init-docker: docker-volumes docker-secrets
+init-docker: docker-volumes docker-secrets hosts-file
```
shaikh-rashid: We would not need to modify hosts file for non development environments (i.e. prod deployments). Therefore it would be preferable to have this as an optional step instead of a default action with init-docker

shaikh-rashid (changes requested):
Minor change requested

shaikh-rashid (approved):
LGTM

daisieh (approved):

PR #198, opened 2023-02-23
Fix the setup_hosts.sh for debian instances on OpenStack

OrdiNeu:
While setting up the federation stuff, I realized that on OpenStack, it uses neither eth0 (ethernet) nor wpl0 (wireless), but instead ens3 (which seems standard for Debian, based on a quick google search).
This PR fixes that, and also moves away from the now-deprecated `ifconfig` in favour of `ip addr`. It also fixes the case where no hosts were found but it was falsely declaring that it found more than 1 IP addresses.

**To test**:
- On a fresh OpenStack instance, running `make init-hosts-file` should work out of the box
- Additionally, on any operating system where the script cannot find the local IP address, it throws an error instead.

### DIFF
--- a/setup_hosts.sh
+++ b/setup_hosts.sh
@@ -11,25 +11,31 @@ fi
 # Replace docker.localhost entry in /etc/hosts
 echo "HACK ALERT"
 echo "This step will introduce an entry into /etc/hosts so that requests to docker.localhost"
-echo "will bounce off your local network. Root access will be required..."
+echo "will bounce off your local network. Root access will be required, and this script assumes"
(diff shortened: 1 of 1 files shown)


daisieh:
Apparently on my Mac (maybe it's bash?) there is no ip command. Maybe check to see if that's present first, and if not, use ifconfig?

daisieh (changes requested):
I left a comment about not having ip on my shell, so if you could keep ifconfig as a fallback, that would be great.

I can't test it on an openstack instance because there are no IPs left in the brudno pool! I believe you that it works, though.

OrdiNeu:
The issue this was fixing is fixed with changes from #203

PR #200, opened 2023-03-08
Conda path fixes

kcranston:
This fixes a few bugs with the conda paths, and make some modifications for M1. Notably, the conda-init now adds the conda path to zsh on M1, so you can activate the env using the name `conda activate candig` and not require the full path to the env. 

To test, run `make bin-conda` and `make init-conda`. You don't need to re-build the entire stack.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,17 @@ export $(shell sed 's/=.*//' $(env))
 
 SHELL = bash
 DIR = $(PWD)
-CONDA_BASE = ${HOME}/bin # this is where miniconda gets installed
-CONDA = $(CONDA_BASE)/miniconda3/bin/conda
-CONDA_ENV_SETTINGS = $(CONDA_BASE)/etc/profile.d/conda.sh
+
+#>>>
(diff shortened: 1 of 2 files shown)


kcranston:
Question for folks - should we move the conda install location to `.env` rather than having it in the Makefile? We need a different location on WSL, and also if you want to use an existing conda install.

SonQBChau:
I think we should have the main branch on Linux because that is the most important, then have Apple branch and WSL branch:
- It's confusing to read both instructions and pick out what relevant
- We have to write `if else` and put comments in the makefile
- It's harder to update, and we can miss out on the instruction
- Each PR requires to be tested on all the machines
- If we don't have anybody on Apple or WSL, we can let that branch out of sync
- We can do whatever we want on the custom branch
- Soon we will have linux on arm, window on arm...We don't have to support them all, but it's easier to make a branch off

yavyx (approved):
Re-built the stack on Linux with conda working properly.

SonQBChau (approved):
worked on M1, can activate using candig name

PR #201, opened 2023-03-09
Make Federation easier to setup

OrdiNeu:
This PR introduces a quick edit to the example `.env` file so that federation works out of the box, and adds a new script to quickly insert new federated instances inside the Tyk config.

**To test**:

- [ ] Inside lib/tyk/configuration_templates, I've included a python script insert-domain.py. If you add in new providers (in the format `domain|secret|policy`, e.g. `http://candig-dev|VgGpzVdAjWCuuJZwNF5IzvJFnm4sWdrc|candig_policy` (not the real secret)), it should add new entries into every .tpl file here.
- [ ] With or without using the script above, you should be able to copy the `.env` file to the root, add `federation-services` to `CANDIG_MODULES` in line 5 of that file, then rebuild the stack. Afterwards, federated calls like the following should work:
```
curl --request POST \
  --url http://docker.localhost:5080/federation/search \
  --header "Authorization: Bearer <$TOKEN>" \
  --header "federated: true" \
  --header 'Content-Type: application/json' \
  --data '{
 "endpoint_payload": {},
  "request_type": "GET",
  "endpoint_path": "moh/v1/version_check",
  "endpoint_service": "katsu"
}'
```

**NB**: You will also need to add the following two files. I didn't know if I should enable this by default.
tmp/federation/servers:
```
{
  "servers": [
    {
      "url": "http://docker.localhost:5080/federation/search",
      "location": [
        "UHN",
        "Ontario",
        "ca-on"
      ]
    }
  ]
}
```

tmp/federation/services:
```
{
  "services": {
    "katsu": "http://docker.localhost:5080/katsu",
    "candig-server": "http://docker.localhost:5080/candig",
    "htsget": "http://docker.localhost:5080/genomics"
  }
}
```

**To test this without taking down the entire stack:**
- `docker stop` any instances of federation you have running
- `make clean-authx`
- `make build-federation-service`
- `make compose-federation-service`
- `make init-authx`


daisieh:
This will be superseded by the big federation refactor: https://candig.atlassian.net/browse/DIG-1131

PR #203, opened 2023-03-14
Bake the `make init_hosts_file` step so it is no longer needed + Add a `make build-all` command

OrdiNeu:
This PR rolls all of the hosts file steps into `extra_hosts` commands on the docker volumes that require it. This means that the `make init-hosts-file` step is no longer used. This PR also includes my `make build-all` command, that should rebuild the entire CanDIG stack from scratch, without needing to go through all of the steps in `docs/install-docker.md`.

**To test**
- There is a new environment variable in your `.env` file, called [LOCAL_IP_ADDR](https://github.com/CanDIG/CanDIGv2/pull/203/files#diff-94bd2ecddc422d4fa0e5cae5238156d394349819728e1d26f1f04f8168618e26R15). You can set this to your local IP address, or the script will try to automatically find the correct IP to use.
- This requires you to remove any edits to your hosts file that you may have set up (i.e. any `docker.localhost` entries in `/etc/hosts`).
- If you want to run the entire new stack, run `make clean-all`, followed by `make init-conda`, `conda activate candig`, `pip install -U -r etc/venv/requirements.txt`, `make build-all`.
- If you just want to test this in the context of authx, run `make clean-authx` followed by `make init-authx`.
- **NB: The `make init-hosts-file` step is no longer necessary and has been removed**


daisieh:
I don't understand the flow here: is setup_hosts supposed to set my /etc/hosts file anymore, or not? If so, it can't, because my system doesn't have `ip` on it, only `ifconfig`. If not, how does the new LOCAL_IP_ADDR and extra_hosts work to set docker.localhost outside of the containers?

If I manually set docker.localhost in /etc/hosts as before, then run make init-authx, everything works as it should, as it always has. If I don't, it fails any time a script tries to call a curl, like when keycloak_setup.sh tries to get the access token, presumably because it can't find docker.localhost. What is extra_hosts supposed to do? And do you need to set it in keycloak's docker-compose.yml as well?

Makefile
```diff
@@ -418,3 +410,21 @@ help:
...
+#<<<
+.PHONY: build-all
+build-all:
+	./pre-build-check.sh
+
+# Setup the entire stack
+	$(MAKE) init-docker
+	$(MAKE) build-minio
+	$(MAKE) build-htsget-server
+	$(MAKE) build-chord-metadata
```
daisieh: You shouldn't have to individually build modules: if you run make compose, they should get pulled or built as needed, based on the CANDIG_MODULES variable in .env.

Makefile
```diff
@@ -418,3 +410,21 @@ help:
...
+	$(MAKE) build-candig-data-portal
+	$(MAKE) build-opa
```
daisieh: Opa gets built/composed in init-authx.

Makefile
daisieh: If you want, you can add the make docker-pull target here instead, so that you get some version pulled down.

Makefile.authx
```diff
@@ -75,6 +75,8 @@ init-authx: mkdir
 	$(MAKE) docker-volumes
 	echo "Setting up Keycloak"; \
 	source ${PWD}/lib/keycloak/keycloak_setup.sh; \
+	echo "Grabbing our local IP address"; \
+	source ${PWD}/setup_hosts.sh; \
```
daisieh, line 81: Why does this happen after the keycloak setup? Keycloak_setup calls curls and expects the /etc/hosts value to be set already.

setup_hosts.sh
```diff
@@ -1,35 +1,33 @@
...
-echo "will bounce off your local network. Root access will be required..."
-cp /etc/hosts .hosts.tmp
-printf "\n" >>.hosts.tmp
-
 if [ "$VENV_OS" == "linux" ]; then
-  ifconfig | grep -A 1 'wlp0\|eth0' | grep -o "inet [0-9.]\+" | cut -d' ' -f2 > .hosts.tmp2
+    ip addr | grep -A 1 'wlp[0-9]\|eth[0-9]\|ens[0-9]' | grep -o "inet [0-9.]\+" | cut -d' ' -f2 > .hosts.tmp2
 else
-  ifconfig | grep -o "inet [0-9.]\+" | cut -d' ' -f2 > .hosts.tmp2
+    ip addr | grep -o "inet [0-9.]\+" | cut -d' ' -f2 > .hosts.tmp2
```
daisieh: My system doesn't have `ip`, only `ifconfig`.

setup_hosts.sh
```diff
@@ -1,35 +1,33 @@
...
+    echo "and add it to /etc/hosts for the 'docker.localhost' host."
+    cat .hosts.tmp2
 else
     echo "ERROR: More than one IP has been detected. Since this script can't automatically determine which one"
-    echo "will make the build work, please remove all but one of the inserted docker.localhost lines in /etc/hosts."
-    awk '{printf "%s\tdocker.localhost\n", $0}' .hosts.tmp2 >> .hosts.tmp
-    sudo mv .hosts.tmp /etc/hosts;
-    rm .hosts.tmp2;
+    echo "will make the build work, please set LOCAL_IP_ADDR=<your local network IP address> (e.g. 192.168.x.x)"
+    echo "in your .env file, then restart conda"
```
daisieh, line 30: If there are more than one IP address available, shouldn't I still need to set one in /etc/hosts as well?

setup_hosts.sh
```diff
@@ -1,35 +1,33 @@
...
-numlines=$(wc -l .hosts.tmp2)
+numlines=$(cat .hosts.tmp2 | wc -l) # use cat to prevent the name of the file from being printed on some systems
 if [ "$numlines" == "1" ]; then
-    printf "$IP\tdocker.localhost" >> .hosts.tmp
-    sudo mv .hosts.tmp /etc/hosts;
-    rm .hosts.tmp2;
+    export LOCAL_IP_ADDR=`cat .hosts.tmp2`
+elif [ "$numlines" == "0" ]; then
+    echo "ERROR: Your internet adapter could not be found automatically. Please determine your local IP address,"
+    echo "and add it to /etc/hosts for the 'docker.localhost' host."
```
daisieh: Does the user not need to set the LOCAL_IP_ADDR value in .env as well here?

daisieh (changes requested):

setup_hosts.sh
OrdiNeu: I'm unable to test this, but would setting `docker.localhost` to `::1` (loopback) work? While testing, I haven't needed to set `docker.localhost` to anything in my /etc/hosts.

OrdiNeu:
@daisieh 
> I don't understand the flow here: is setup_hosts supposed to set my /etc/hosts file anymore, or not? If so, it can't, because my system doesn't have `ip` on it, only `ifconfig`. If not, how does the new LOCAL_IP_ADDR and extra_hosts work to set docker.localhost outside of the containers?

The `setup_hosts` script no longer sets your /etc/hosts (and you shouldn't need to touch it anymore besides maybe assigning it to loopback (::1). I just set a patch for the `ip` thing; [ifconfig has been deprecated and is no longer installed by default on newer linux machines,](https://www.redhat.com/sysadmin/ifconfig-vs-ip) so I have to use it for e.g. debian instances on OpenStack.

> If I manually set docker.localhost in /etc/hosts as before, then run make init-authx, everything works as it should, as it always has. If I don't, it fails any time a script tries to call a curl, like when keycloak_setup.sh tries to get the access token, presumably because it can't find docker.localhost. What is extra_hosts supposed to do? And do you need to set it in keycloak's docker-compose.yml as well?

The command sets the value of `docker.localhost` _inside the containers_ themselves, such that you can get the following:
```
(candig) fnguyen@fnguyen-worktop:~/Documents/CanDIGv2$ docker exec -it candigv2_opa-runner_1 bash
bash-5.1# cat /etc/hosts
127.0.0.1       localhost
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
10.141.12.207   docker.localhost
172.18.0.14     3a287ad64ffc
```

The end result is that you don't need to edit your `/etc/hosts` file anymore (besides maybe setting it to `::1`?). In essence, this is one step you need to run to bring up CanDIG. As for why it wasn't in Keycloak -- my setup appeared to work without it, but I can also add it ([722412f](https://github.com/CanDIG/CanDIGv2/pull/203/commits/722412ff3b4907b042d6e458559bb4550b902c84)) to avoid anything going awry

daisieh:
> The end result is that you don't need to edit your `/etc/hosts` file anymore (besides maybe setting it to `::1`?). In essence, this is one step you need to run to bring up CanDIG. 

I should add it like:
```
::1   docker.localhost
```
?

OrdiNeu:
> ```
> ::1   docker.localhost
> ```
Yes, either that or 127.0.0.1 :+1:

daisieh:
I've added a few changes that I needed to make for my setup that will probably not hurt anyone else's setup in #205, if you want to review/merge that into this one. With those changes and the loopback (probably worth putting in the readme or docs or whatever), this all works fine for me now!

setup_hosts.sh
daisieh: Yes, this works for me, and is obviously way easier and more robust than having to update my actual IP when I go from ethernet to wifi.

daisieh (approved):

daisieh:
I approved this PR, but forgot to remind that I would like there to be a note somewhere in the docs about loopback IP, because I didn't know about that and didn't have that set up yet.

PR #206, opened 2023-03-29
Fix a bug where make compose-% was missing the IP address

OrdiNeu:
Bugfix a regression introduced by #203 where the `make compose-%` command was failing due to the missing environment variable.

**To test**: Attempt to bring a module down, then back up with `make compose-candig-data-portal` (for instance)


CourtneyGosselin (approved):
LGTM!

PR #207, opened 2023-03-31
DIG-999: Automate integration tests

daisieh:
Integration testing is based on the overall existence of datasets SYNTHETIC-1 and SYNTHETIC-2 that are built into our Opa setup. Test data is installed for katsu and htsget, but not uninstalled.

(You might need to pull https://github.com/CanDIG/candig-opa/pull/32 before running these tests)

To run a basic suite of authorization tests, based on the [Confluence page](https://candig.atlassian.net/wiki/spaces/CA/pages/665518081/Integration+testing), run `make test-integration`.

I removed a lot of old tests that we haven't used in years and that I think this can replace.

Please review the tests that I chose to write for your features of interest and make sure that they do reasonable things and that I've tested enough basic functionality. Federation isn't covered here because it needs some reworking first.


etc/tests/test_integration.py
```diff
@@ -0,0 +1,306 @@
...
+        'Authorization': f"Bearer {site_admin_token}",
+        'Content-Type': 'application/json; charset=utf-8'
+    }
+    response = requests.post(f"{ENV['CANDIG_URL']}/katsu/moh/v1/ingest/programs", headers=headers, json=response.json())
+    print(response.json())
+    if response.status_code >= 400:
+        errors = response.json()['error during ingest_programs']
+        assert "code='unique'" in errors and "program_id" in errors # this means that the error was just that the program IDs already exist
+    else:
+        assert response.status_code == 204 or response.status_code == 200
```
SonQBChau: this should be 201 CREATED, assert 204 or 200 will fail

etc/tests/test_integration.py
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    response = requests.get(f"{ENV['CANDIG_URL']}/katsu/moh/v1/authorized/programs/", headers=headers)
+    programs = map(lambda x: x['program_id'], response.json()['results'])
+    print(programs)
+    assert dataset in programs
+    assert not_dataset not in programs
+
+    response = requests.get(f"{ENV['CANDIG_URL']}/katsu/moh/v1/authorized/donors/", headers=headers)
+    assert len(response.json()) > 0
```
SonQBChau, line 299: This assert fails since len = 0. I don't see donors being ingested, can you double-check?

SonQBChau (changes requested):

etc/tests/test_integration.py
daisieh: oops, forgot to ingest donors! fixed.

SonQBChau:
I still got :
```
______________________________________________________________ test_setup_katsu ______________________________________________________________

    def test_setup_katsu():
        test_loc = "https://raw.githubusercontent.com/CanDIG/katsu/develop/chord_metadata_service/mohpackets/data/small_dataset/synthetic_data/Program.json"
        response = requests.get(test_loc)
        assert response.status_code == 200
    
        site_admin_token = get_token(username=ENV['CANDIG_SITE_ADMIN_USER'], password=ENV['CANDIG_SITE_ADMIN_PASSWORD'])
        headers = {
            'Authorization': f"Bearer {site_admin_token}",
            'Content-Type': 'application/json; charset=utf-8'
        }
        response = requests.post(f"{ENV['CANDIG_URL']}/katsu/moh/v1/ingest/programs", headers=headers, json=response.json())
        print(response.json())
        if response.status_code >= 400:
            errors = response.json()['error during ingest_programs']
            assert "code='unique'" in errors and "program_id" in errors # this means that the error was just that the program IDs already exist
        else:
>           assert response.status_code <= 200 and response.status_code < 300
E           assert (201 <= 200)
E            +  where 201 = <Response [201]>.status_code

etc/tests/test_integration.py:261: AssertionError
```

daisieh:
fixed.

SonQBChau (approved):
tested. LGTM!

PR #208, opened 2023-03-31
DIG-1018, DIG-1019, DIG-1020, DIG-1021: fix make clean targets

daisieh:
I labeled all of our containers/volumes/secrets as they are created and so we can filter on those labels before we delete or remove them on clean.

NOTE: you still have to delete all of the unlabeled containers/volumes and re-create them before they'll be labeled, so if you couldn't use make clean-all before, you'll have to manually delete the candigv2 ones and then run make compose to recreate them.

I think that the bug listed in DIG-1017 is slightly different and might've been addressed by https://github.com/CanDIG/CanDIGv2/pull/203, but testing will show if that's true.

PR #210, opened 2023-04-04
Label PostGres + Vault volumes

OrdiNeu:
This PR labels one of the volumes that were discovered to be unlabeled during testing of #208 -- one from Postgres. There's two more from Vault one that I cannot specify ([/vault/file and /vault/logs](https://hub.docker.com/_/vault)) that I cannot specify without the vault setup going awry.

The effect of this PR is that when you run `make compose; make init-authx`, there is one less unlabelled volumes created, and these will be properly deleted with `make clean-all`.


daisieh (approved):
Works for me!

PR #213, opened 2023-04-05
Test htsget inside its container

daisieh:
This PR includes changes to htsget to allow one to run pytest inside the container.

To test:
* Rebuild htsget (`make build-htsget-server`, `make compose-htsget-server`)
* Run pytest inside the container (`docker exec -it candigv2_htsget-app_1 pytest`)
* Verify that pytest still works outside the container (`pytest lib/htsget-server/htsget_app/tests/`)


OrdiNeu (approved):
Tests, integration tests, and running from within or outside the container all work. LGTM :heavy_check_mark:

PR #216, opened 2023-04-19
DIG-1028: standardize preflight/setup files in compose targets

daisieh:
I decided that the best approach would be to have two available default scripts:
* %_preflight.sh runs before the docker compose step, to prepare any files that need to be moved to the container
* %_setup.sh runs after the container is created, to do any setup needed after the container exists

The make compose-% target calls them both, but doesn't fail if the scripts don't exist. I removed all other references to setup scripts in init-authx.

I created template versions in the templates directory and added some info to that README.

You can test to make sure everything is the same as before:
```
make clean-authx
make init-authx
make test-integration
```

(katsu tests might still be failing, but that is unrelated.)

I also deleted an unused setup_auth.sh script, took out an unnecessary line in htsget's docker-compose, and added a diff line to pre-build-check.sh.


SonQBChau:
I got this error when running make build-all:
```
./pre-build-check.sh
./pre-build-check.sh: line 3: getent: command not found
```

OrdiNeu:
Per standup discussion yesterday, the diff line prevented people whose .env differed from example.env from continuing with the installation process (and would not be noticeable without a prompt otherwise). On suggestions, this functionality should be moved to a different PR.

SonQBChau:
tested and worked for me

SonQBChau (approved):

PR #217, opened 2023-04-20
Fix an issue with pre-build-check which caused it to fail on certain MacOS systems

OrdiNeu:
Fixing an issue where the `make build-all` command fails on certain MacOS systems because getent doesn't work on them.


SonQBChau:
tested and worked

SonQBChau (approved):

PR #221, opened 2023-04-25
Remove diff flag not available on all versions

daisieh:


SonQBChau (approved):
Worked!
Can you make a small change to add a yes/no at the prompt? It took me 3 tries since I didn't know the expected input.
```
read -r -p 'Do you want to continue? (y/n)' choice
```

PR #222, opened 2023-04-28
Rename htsget-server/htsget-app to htsget

daisieh:
This will make the service and container name more consistent with the names in the .env file.

Test:
```
make clean-all
git checkout daisieh/htsget-rename
rm -R lib/htsget-server/
cd lib/htsget/htsget_app/
git submodule init
git submodule update --recursive
make bin-conda
make build-all
make test-integration
```


OrdiNeu (approved):
Works on my machine, the `test_keycloak` function is a good addition :heavy_check_mark:

PR #227, opened 2023-05-09
Add pre-build-checks to prevent the problems found during BCGSC test server deployment

OrdiNeu:
This adds two more checks:
1. That the submodules have been checked out
2. That CANDIG_DOMAIN can be reached, whatever its value is


daisieh (approved):
I added a suggestion for a change in https://github.com/CanDIG/CanDIGv2/pull/228; otherwise everything looks good and works as expected.

PR #230, opened 2023-05-15
Make makefiles safer

justin-ys:
- Makefiles previously set a variable "DIR" to the present working directory and used this to run commands. In rare cases, it seems like this may resolve to an empty string. This can be very dangerous, especially for clean commands. This commit changes the paths to be relative instead.


daisieh (approved):
Works for me, thanks!

PR #233, opened 2023-05-16
DIG-1166: reorder vault setup

daisieh:
It looks like the problem is that the Dockerfile writes things to /vault, which is an external volume that may not exist yet. I reordered the setup steps so that Dockerfile loads everything into /app, the local workdir, and then entrypoint copies it to /vault as part of the initial setup.

To test:
```
make clean-authx
make init-authx
make test-integration
```

### DIFF
--- a/lib/vault/Dockerfile
+++ b/lib/vault/Dockerfile
@@ -17,13 +17,11 @@ RUN apk add --no-cache \
 WORKDIR /app
 
 # add the config file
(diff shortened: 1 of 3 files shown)


SonQBChau (approved):
Worked on mine! I also tested with my updated subrepos and change the test-integration and they all work!

justin-ys (approved):
Works on my machine.

PR #234, opened 2023-05-19
DIG-1143: test SampleDrsObjects

daisieh:
Picked up a few fixes that I noticed about drs self-uris, added a test for creating a sampledrsobject. `make test-integration` should pass after you've picked up and rebuilt htsget from https://github.com/CanDIG/htsget_app/pull/261.


OrdiNeu (approved):
Integration tests pass on my machine. Code looks clean. :+1:

PR #235, opened 2023-05-24
DIG-1172: Build validation

justin-ys:
- Addresses [DIG-1172](https://candig.atlassian.net/browse/DIG-1172)
- Adds an automatically running script to validate the number of running containers based on .env after build
- Logs stderr and any lines from stdout containing "error" or "warning" (with 2 lines of context above/below) into a log file (default: tmp/error.txt)
- Note this removes color from docker building/composing since it switches to "plain" mode when piping to tee.

To test:
- Add ERRORLOG=tmp/error.txt to the end of your .env.
- Run make build-all.
- Ensure there is some output in tmp/error.txt.
- If the build was successful, ensure that a success message in green was printed at the end of the build. If not, a warning message should have appeared. If so, make sure the number of containers it reports are currently running and number of containers it states should be running are correct.

[DIG-1172]: https://candig.atlassian.net/browse/DIG-1172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,6 @@ CONDA = $(CONDA_INSTALL)/miniconda3/bin/conda
 CONDA_ENV_SETTINGS = $(CONDA_INSTALL)/miniconda3/etc/profile.d/conda.sh
 
 LOGFILE = tmp/progress.txt
-ERRORLOG = tmp/error.txt
 
 $(shell printf "Build started at `date '+%D %T'`.\n\n" >> $(ERRORLOG) $(ERRORLOG))
 
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,7 @@ build-%:
 	printf "\n\nErrors during build-$*: \n" >> $(ERRORLOG)
 	echo "    started build-$*" >> $(LOGFILE)
 	DOCKER_BUILDKIT=1 COMPOSE_DOCKER_CLI_BUILD=1 \
-	docker compose -f lib/candigv2/docker-compose.yml -f lib/$*/docker-compose.yml build $(BUILD_OPTS) --progress=plain 2> >(tee -a $(ERRORLOG))
+	docker compose -f lib/candigv2/docker-compose.yml -f lib/$*/docker-compose.yml build $(BUILD_OPTS) 1> >(tee -a >(grep -C 2 "[Ee]rror|[Ww]arning" >> $(ERRORLOG)) /dev/null) 2> >(tee -a $(ERRORLOG))
 	echo "    finished build-$*" >> $(LOGFILE)
 
 
@@ -241,10 +241,9 @@ compose:
 compose-%:
 	printf "\n\nErrors during compose-$*: \n" >> $(ERRORLOG)
 	echo "    started compose-$*" >> $(LOGFILE)
-	-source lib/$*/$*_preflight.sh 2> >(tee -a $(ERRORLOG))
-	source setup_hosts.sh; \
-	docker compose -f lib/candigv2/docker-compose.yml -f lib/$*/docker-compose.yml --compatibility up -d 2> >(tee -a $(ERRORLOG))
-	-source lib/$*/$*_setup.sh 2> >(tee -a $(ERRORLOG))
+	-source lib/$*/$*_preflight.sh 1> >(tee -a >(grep -C 2 "[Ee]rror|[Ww]arning" >> $(ERRORLOG)) /dev/null) 2> >(tee -a $(ERRORLOG))
+	docker compose -f lib/candigv2/docker-compose.yml -f lib/$*/docker-compose.yml --compatibility up -d 1> >(tee -a >(grep -C 2 "[Ee]rror|[Ww]arning" >> $(ERRORLOG)) /dev/null) 2> >(tee -a $(ERRORLOG))
+	-source lib/$*/$*_setup.sh 1> >(tee -a >(grep -C 2 "[Ee]rror|[Ww]arning" >> $(ERRORLOG)) /dev/null) 2> >(tee -a $(ERRORLOG))
 	echo "    finished compose-$*" >> $(LOGFILE)
 
 
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ CONDA_ENV_SETTINGS = $(CONDA_INSTALL)/miniconda3/etc/profile.d/conda.sh
 LOGFILE = tmp/progress.txt
 ERRORLOG = tmp/error.txt
 
-$(shell > $(ERRORLOG))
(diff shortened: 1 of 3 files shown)


post_build.sh
```diff
@@ -0,0 +1,37 @@
+RED='\033[0;31m'
```
OrdiNeu, line 11:
```suggestion
#!/usr/bin/env bash

RED='\033[0;31m'
```
Opening shebang

post_build.sh
```diff
@@ -0,0 +1,37 @@
...
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+GREEN='\033[0;32m'
+DEFAULT='\033[0m'
+
```
OrdiNeu, line 16: This script works great! Given the name of it though (`post_build.sh`), a comment explaining what the rest of the script is doing would be nice, since it is possible that we might extend it to do more in the future. See [https://github.com/CanDIG/CanDIGv2/blob/develop/pre-build-check.sh](here) for an example of a script that we ended up extending in the future -- it would be nice to have a comment here saying that we're making sure the correct number of containers have been initialized.

Makefile
```diff
@@ -234,11 +242,12 @@ compose:
 
 #<<<
 compose-%:
+	printf "\n\nErrors during compose-$*: \n" >> $(ERRORLOG)
 	echo "    started compose-$*" >> $(LOGFILE)
-	-source lib/$*/$*_preflight.sh
+	-source lib/$*/$*_preflight.sh 1> >(tee -a >(grep -C 2 "[Ee]rror|[Ww]arning" >> $(ERRORLOG)) /dev/null) 2> >(tee -a $(ERRORLOG))
```
OrdiNeu: Prefer `-i` to including both capitalizations like this, since all-caps error/warning messages are also common. In addition, warnings often have just `WARN` in them.
```suggestion
	-source lib/$*/$*_preflight.sh 1> >(tee -a >(grep -iC 2 "error|warn" >> $(ERRORLOG)) /dev/null) 2> >(tee -a $(ERRORLOG))
```
I'm also unsure about storing only the lines matching error/warn here -- it might be best to store *everything*, and then grep for errors afterwards (during post-build).

post_build.sh
```diff
@@ -0,0 +1,37 @@
...
+	SERVICE_COUNT=$((SERVICE_COUNT + MODULE_SERVICES))
+done
+
+if [ $(docker ps -q | wc -l) == $SERVICE_COUNT ]
+then
+	echo -e "${GREEN}Number of expected CanDIG services matches number of containers running!${DEFAULT}"
+else
+	echo -e "${RED}WARNING: ${YELLOW}The number of CanDIG containers running does not match the number of expected services.\nRunning: ${BLUE}$(docker ps -q | wc -l) ${YELLOW}Expected: ${BLUE}${SERVICE_COUNT}
+${DEFAULT}Check your build/docker logs."
+fi
```
OrdiNeu, line 86: Might be nice to also print out all of the errors that have occurred (e.g. print out tmp/error.txt or grep out relevant errors) during this step, since currently our problem is that the errors whiz by the screen without being easily readable.

Makefile
```diff
@@ -121,9 +127,11 @@ build-images: #toil-docker
 
 #<<<
 build-%:
+	printf "\n\nErrors during build-$*: \n" >> $(ERRORLOG)
 	echo "    started build-$*" >> $(LOGFILE)
+	source setup_hosts.sh; \
 	DOCKER_BUILDKIT=1 COMPOSE_DOCKER_CLI_BUILD=1 \
-	docker compose -f lib/candigv2/docker-compose.yml -f lib/$*/docker-compose.yml build $(BUILD_OPTS)
+	docker compose -f lib/candigv2/docker-compose.yml -f lib/$*/docker-compose.yml build $(BUILD_OPTS) 1> >(tee -a >(grep -C 2 "[Ee]rror|[Ww]arning" >> $(ERRORLOG)) /dev/null) 2> >(tee -a $(ERRORLOG))
```
OrdiNeu: I'm noticing a lot of unintended behaviour with tee -- notably, that it messes up coloured output, and also with download progress. Is there an alternative we can look into?

OrdiNeu (changes requested):
Good job! I have a few suggestions & questions -- please review and implement what you can.

In addition, I couldn't track it down, but something is overwriting the tmp/error.txt file instead of appending to it, causing it to get wiped quite often. Could we reserve that behaviour to the `clean-all` step or similar?

Makefile
justin-ys: Unfortunately tee does change the output format, though not directly - Docker, in detecting that it is piping to tee, will "simplify" its output so it can be written to a file. Otherwise, if color codes and other formatting were kept, logs would look like [this](https://imgur.com/ml48nfL). I don't think an alternative to tee would exactly work since the root problem is that Docker's "regular" output simply doesn't work nicely with files. One possible solution I just saw online is manually filtering out ANSI escape codes using sed, which would make logs more readable, but this would make the logging lines very bulky indeed. I will look into whether they can be put into a function again.

justin-ys:
> In addition, I couldn't track it down, but something is overwriting the tmp/error.txt file instead of appending to it, causing it to get wiped quite often. Could we reserve that behaviour to the clean-all step or similar?

I put in a line to clean the Error log every time it is run (`$(shell > $(ERRORLOG))`). Maybe I can just append a new datetime string as a separator instead?

post_build.sh
justin-ys: The post build script will now print out lines only containing "error" or "warn" for each module, but the Makefile will store all output from stdout/stderr into the error log.

post_build.sh
```diff
@@ -0,0 +1,80 @@
...
+if [ $(docker ps -q | wc -l) == $SERVICE_COUNT ]
+then
+	echo -e "${GREEN}Number of expected CanDIG services matches number of containers running!${DEFAULT}"
+else
+	RUNNING_MODULES=$(docker ps --format "{{.Names}}")
+	for MODULE in $ALL_MODULES; do
+		printf "\n\n${RED}Error logs for ${MODULE}:\n--------------------\n${DEFAULT}"
+		print_module_logs $MODULE
+		printf "${RED}--------------------\n${DEFAULT}"
+	done
```
OrdiNeu, line 83: This works great! Only request now is to move this outside of the `if` statement, as I think it would be helpful to always see all errors/warnings that have occurred. Besides that, everything looks great!

OrdiNeu (changes requested):

OrdiNeu (approved):

PR #238, opened 2023-05-30
When assertions fail, print response.text

daisieh:
This way, we can more quickly see what sorts of failures we got.


daisieh:
Merged into #239

PR #243, opened 2023-06-15
DIG-921: Fix the "Websocket error: Could not connect to ws:// <x>" error

OrdiNeu:
[Jira link](https://candig.atlassian.net/browse/DIG-921)

This fixes the candig-data-portal error that would spam the console with `WebSocket connection to 'ws://candig.docker.internal:3000/ws' failed:`.

In addition, this makes linter/prettier errors show up when browsing the web page. Unsure if this is functionality we want, but I found it pretty useful as a reminder.


CourtneyGosselin (approved):
LGTM works as expected

PR #248, opened 2023-07-11
GitHub Actions for CanDIGv2

justin-ys:
I wanted to finish implementing workflow dispatches before this PR but unfortunately I need to commit the action to the main branch at least once first; I think it should be finished regardless but unfortunately I may need to make another PR if it requires fixes.

This PR creates a GitHub action which builds the CanDIGv2 stack, ensures all containers are running, and runs integration tests. It is triggered on PRs/commits to main as well as (if working) commits to the main repositories of submodules. Build and container logs are saved as an artifact within the repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,12 +93,17 @@ build-all:
 
 # Setup the entire stack
 	$(MAKE) init-docker
-	$(MAKE) init-conda
 	$(MAKE) build-images
 	$(MAKE) compose
 	$(MAKE) init-authx
 	
 	./post_build.sh
+	
+.PHONY: install-all
+install-all:
+	$(MAKE) bin-conda
+	$(MAKE) init-conda
+	$(MAKE) build-all
 
 
 #>>>
--- a/Makefile
+++ b/Makefile
@@ -15,9 +15,8 @@ SHELL = bash
 #  and then use make mkdir and make init-conda (no bin-conda, which will blow up an existing conda)
 # <<<
 
-CONDA_INSTALL = bin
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -15,9 +15,8 @@ SHELL = bash
 #  and then use make mkdir and make init-conda (no bin-conda, which will blow up an existing conda)
 # <<<
 
-CONDA_INSTALL = bin
```
daisieh, line 18: Is removing this line going to be a problem for other platforms?

daisieh: Oh I see, it moved to .env

etc/tests/test_integration.py
```diff
@@ -163,6 +163,7 @@ def test_vault():
 
     # confirm that this works with the CANDIG_S3_TOKEN too:
     headers["X-Vault-Token"] = ENV["VAULT_S3_TOKEN"]
+    print(f"vault token is {headers['X-Vault-Token']}")
```
daisieh: This can probably be removed...

daisieh (changes requested):
Can you do a review of https://github.com/CanDIG/CanDIGv2/pull/247 so that we can merge that one first? Then you can rebase this branch on develop and we can keep this PR clean.

etc/tests/test_integration.py
justin-ys: Resolved by [639c87b](https://github.com/CanDIG/CanDIGv2/pull/248/commits/639c87b8b1255f57ab21b2fa1b28c05d70b20a50)

daisieh (approved):
Looks good!

PR #249, opened 2023-07-18
Conda hotfix

justin-ys:
- Fixes conda installation? (previously, incorrect paths were used for bin/init-conda, so it would not run)
- To test: make clean-all, make install-all


SonQBChau:
`make install-all` doesn't work on m1
`make bin-all` + `make init-conda` + `conda activate` candig worked, passed test-integration

SonQBChau (approved):

PR #251, opened 2023-07-21
Update the candig-data-portal ref

OrdiNeu:


daisieh (approved):
Looks good, passes integration tests.

PR #250, opened 2023-07-20
Sonchau/add katsu secret

SonQBChau:
What's new:
- add function to make katsu key (50 chars)
- generate katsu secret
- add katsu prod yaml
- add clean up test in test-integration, now katsu should clean itself any extra data after ingest

How to test:
- update katsu to sonchau/part_36
- make test-integration

PR #253, opened 2023-07-21
Update katsu, fix integration tests

SonQBChau:
this update katsu submodule to fix the integration test failure from #250


daisieh (approved):
Works on my machine; I had to specifically run `make katsu-secrets` before re-composing katsu, but then after that, integration tests passed.

PR #270, opened 2023-08-01
DIG-1273: Add more endpoints to the Tyk test, and add a timeout

OrdiNeu:
[Jira ticket](https://candig.atlassian.net/browse/DIG-1273)

This PR makes sure Tyk is able to communicate with all endpoints, rather than just HTSGet.

**To test**:
Setup the entire stack, as usual, then:
1. Check that the Test_Tyk works if the entire setup is working, and then:
2. Fails if either OPA, Federation, HTSGet, or Katsu are down

You can run this specific test with `python settings.py; source env.sh; pytest ./etc/tests -k "Test_Tyk"`, or by running all integration tests (make test-integration).


daisieh (changes requested):
Maybe add a print so you can see which endpoint failed?

OrdiNeu:
@daisieh 
> Maybe add a print so you can see which endpoint failed?

Fixed :heavy_check_mark:

OrdiNeu:
Unsure what's causing that build failure (conda failing to install??), checking...

daisieh (approved):
Looks good!

OrdiNeu:
No idea why the check failed again -- it's failing on a completely different test. Going to try retriggering

PR #271, opened 2023-08-11
Prod changes that still work on local installs

OrdiNeu:
This PR consists of a few changes that allow the setup to work with production instances as well as local instances.
Notably, this is meant to only include the changes that work with both production and your local install -- it should otherwise not affect the installation process for local installs at all.

** To test: **
- Build & make integration tests. Nothing should be broken with the introduction of these changes.

### DIFF
--- a/lib/federation/initialize.py
+++ b/lib/federation/initialize.py
@@ -41,22 +41,22 @@ def main():
     headers["Authorization"] = f"Bearer {token}"
 
     print("Adding servers to federation...")
-    url = f"{get_env_value('FEDERATION_SERVICE_URL')}/federation/v1/servers"
+    url = f"http://localhost:{get_env_value("FEDERATION_PORT")}/federation/v1/servers"
     response = requests.request("POST", url, headers=headers, json=server)
     # add other federated servers here
     if response.status_code != 200:
         print(f"POST response: {response.status_code} {response.text}")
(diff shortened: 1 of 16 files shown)


lib/federation/initialize.py
```diff
@@ -41,22 +41,22 @@ def main():
...
 
     print("Adding servers to federation...")
-    url = f"{get_env_value('FEDERATION_SERVICE_URL')}/federation/v1/servers"
+    url = f"http://localhost:{get_env_value("FEDERATION_PORT")}/federation/v1/servers"
     response = requests.request("POST", url, headers=headers, json=server)
     # add other federated servers here
     if response.status_code != 200:
         print(f"POST response: {response.status_code} {response.text}")
-    url = f"{get_env_value('FEDERATION_SERVICE_URL')}/federation/v1/servers"
+    url = f"http://localhost:4232/federation/v1/servers"
```
daisieh: What is this bit for?

OrdiNeu: FEDERATION_SERVICE_URL is not resolvable on our prod instance (candig.uhnresearch.ca exposes only port 80, and proxies requests to port 5080). Instead, I figured localhost would always resolve properly. Come to think of it though, I could use FEDERATION_PORT instead of hardcoding 4232.

daisieh (approved):
I fixed the formatting strings in initialize.py in federation, but otherwise everything looks fine and works well!

PR #272, opened 2023-08-15
DIG-1282: Fix docker-compose dependency issues & GitHub Actions

justin-ys:
- Removes docker-compose from the Conda install requirements
- Fixes the Conda install location for GitHub Actions
    - NOTE: This will require a new CONDA_INSTALL_LOCATION value in your .env, see the example.env


OrdiNeu (approved):
Looks good to me. As long as the actions go through clean this will be good to merge :heavy_check_mark:

PR #278, opened 2023-08-18
Add ingest to the Tyk configuration

OrdiNeu:
**This requires a change to the .env file**

# Description
This adds the candig ingest microservice to Tyk, allowing requests to ${CANDIG_URL}/ingest to point to the ingest microservice, and transforming Keycloak cookies to bearer tokens.

# To test
- All things accessible through candig.docker.internal:1235 should also be accessible through candig.docker.internal:5080/ingest
- Authenticated requests should work properly as well

### DIFF
--- a/etc/env/example.env
+++ b/etc/env/example.env
@@ -124,7 +124,7 @@ FEDERATION_SERVICES=katsu htsget
 define FEDERATION_SELF_SERVER="
 {
     'id': 'internal-1',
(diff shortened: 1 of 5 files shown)


justin-ys (approved):
Works, note that candig.docker.internal:5080/ingest redirects to candig.docker.internal:5080/service-info due to incorrect routing in the microservice. But this isn't related to the configuration here so it will have to be fixed in a future candigv2-ingest PR.

PR #280, opened 2023-08-22
Support for Data Portal ingest page

justin-ys:
- Prepares repository for https://github.com/CanDIG/candig-data-portal/pull/94 (changes are non-breaking, this will work with the old data portal)
- Tyk will convert cookies to access tokens instead of ID tokens
- Added variable in Data Portal docker compose for ingest URL


mshadbolt:
Making this into a draft PR as it will probably need quite a bit of work to bring it up to speed with the stack before merging

PR #288, opened 2023-09-15
Update services and components list in README

mshadbolt:
- Update services and components table
- rearrange order to have services up front
- fixed a typo and added code type formatting
- removed `weavescope logging monitoring` from the deployment instructions because I don't believe they are used anymore, should these also be removed from the `lib` directory? Then again I also believe that we will reinstate them soon, maybe next sprint so maybe not worth fully removing

Reviewer to check the list to see whether anything is missing and the descriptions seem to make sense.

I rearranged ordering to have the services upfront but open to it moving back lower depending on what people think. Possibly thought having the list up front is a handy index.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Optional modules follow the `#` and include federation service, various monitori
 
 For federated installations, you will need `federation-service`.
 
-For production deployments, you will probably want to include  `federation-service`. Be aware that the last three require more resources, includeing storage.
+For production deployments, you will probably want to include  `federation-service`.
 
 Authorization and authentication modules defined in  `CANDIG_AUTH_MODULES` are only installed if you run `make init-authx` during deployment.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,13 +93,15 @@ Not all CanDIG modules are required for a minimal installation. The `CANDIG_MODU
...
+```
   CANDIG_MODULES=minio htsget-server katsu candig-data-portal
+```
 
 Optional modules follow the `#` and include federation service, various monitoring components, workflow execution, and some older modules not generally installed.
 
 For federated installations, you will need `federation-service`.
 
-For production deployments, you will probably want to include  `federation-service weavescope logging monitoring`. Be aware that the last three require more resources, includeing storage.
+For production deployments, you will probably want to include  `federation-service`. Be aware that the last three require more resources, includeing storage.
```
kcranston, line 104: If we are taking out the three monitoring services, then we can also remove the sentence starting with "Be aware..."

kcranston: If we are taking out the three monitoring services, then we can also remove the sentence starting with "Be aware..."

kcranston (changes requested):
Thanks for this! One minor comment, but otherwise LGTM.

README.md
mshadbolt: Thanks good catch! I have updated that in the latest commit.

kcranston (approved):

PR #290, opened 2023-09-19
remove alpine variable from htsget compose

kcranston:
tiny PR to remove the alpine version from the htsget compose file


daisieh (approved):
Yep, makes sense.

PR #293, opened 2023-09-26
DIG-1305: Update candigv2-ingest

daisieh:
I created this manual submodule PR just to see if the GitHub action runs faster.


OrdiNeu (approved):
As written on the other review, `/etc/debian_version` shows up properly :heavy_check_mark:

PR #302, opened 2023-10-19
DIG-1342: update ingest to cleaned-up version

daisieh:
The ingest docker-compose had a couple of environment variables that should be set through the .env file itself (and the secrets), not relying on env.sh having been run.

In addition, note that in example.env, I've moved candig-ingest to the authx modules, because it is reliant on keycloak already having been set up.

To test, log into the docker container:
```
CanDIGv2$ docker exec -it candigv2_candig-ingest_1 /bin/bash
root@1a8b4d5f3032:/ingest_app# export CANDIG_SITE_ADMIN_PASSWORD=<value of your lib/tmp/secrets/keycloak-test-user2-password>
root@1a8b4d5f3032:/ingest_app# export CANDIG_SITE_ADMIN_USER=user2
root@1a8b4d5f3032:/ingest_app# export CLINICAL_DATA_LOCATION=/ingest_app/single_ingest.json 
root@1a8b4d5f3032:/ingest_app# python katsu_ingest.py 
Validating input
Validating the mapped schema...
Validation success.
Beginning ingest
...
```

### DIFF
--- a/etc/env/example.env
+++ b/etc/env/example.env
@@ -2,8 +2,8 @@
 # - - -
 
 # site options
-CANDIG_MODULES=minio postgres htsget katsu candig-data-portal candig-ingest query #drs-server wes-server logging monitoring
-CANDIG_AUTH_MODULES=keycloak tyk opa vault federation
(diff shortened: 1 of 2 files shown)


yavyx (approved):
Tested and working

PR #310, opened 2023-10-27
Yavyx/update installation guide

yavyx:
Added Ubuntu-specific instructions to install Docker.


docs/install-docker.md
```diff
@@ -52,7 +55,41 @@ sudo systemctl start docker
 sudo usermod -aG docker $(whoami)
 ```
 
-Note that CanDIG requires **Docker Compose v2**, which is provided alongside the latest version of Docker. Versions of Docker which do not provide Docker Compose will unfortunately not work with CanDIG.
```
mshadbolt, line 55: Is this note no longer relevant?

yavyx: It's still there, I moved it to the top section where it makes more sense.

mshadbolt: oh right sorry I missed that!

mshadbolt (approved):
Looks good to me!

PR #312, opened 2023-10-27
DIG-1301: delete cascades, rename htsget dataset to cohort

daisieh:
Integration tests should check this. It depends on https://github.com/CanDIG/htsget_app/pull/274 and then tests should pass.

I also tweaked the crontab for vault.


OrdiNeu (approved):
Works on my end, looks good.

PR #319, opened 2023-11-02
DIG-1171: merge Makefile.authx into Makefile

daisieh:
As per DIG-1171:
* I made a new clean-% target that removes the target-named volumes and shuts down the container(s), as well as removing any temp files.
* The new clean-% target replaces specific authx clean targets. After removing unused targets in Makefile.authx, I could then move all remaining targets into Makefile and delete Makefile.authx.
* I also realized that the preflight script does things that are needed for building the image, so I moved that to build-% instead of compose-%.
* Because so many of the auth targets are dependent on the successful building and composing of the others, I rearranged init-authx so that each module is built and composed in sequence, instead of being all built and then all composed.
* I added some checks in keycloak_setup.sh that are described in 1171. Now if you run `make compose-keycloak` twice, it will quit early the second time with an error.

`make clean-authx` and `make init-authx` should work as before, but you should also be able to `make clean-` any module you want.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,6 @@
 env ?= .env
 
 include $(env)
-include Makefile.authx
 export $(shell sed 's/=.*//' $(env))
 
 SHELL = bash
--- a/Makefile
+++ b/Makefile
@@ -373,6 +373,18 @@ docker-volumes:
 
 
 #>>>
+# authx, common settings
+# make init-authx
+
+#<<<
+.PHONY: init-authx
+init-authx: mkdir
+	$(MAKE) docker-volumes
+	$(foreach MODULE, $(CANDIG_AUTH_MODULES), $(MAKE) build-$(MODULE);)
+	$(foreach MODULE, $(CANDIG_AUTH_MODULES), $(MAKE) compose-$(MODULE);)
+
+
+#>>>
 # initialize conda environment
 # make init-conda
 
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,8 @@ build-images: #toil-docker
 build-%:
 	printf "\nOutput of build-$*: \n" >> $(ERRORLOG)
 	echo "    started build-$*" >> $(LOGFILE)
-	source setup_hosts.sh; \
+	source setup_hosts.sh
+	-source lib/$*/$*_preflight.sh 2>&1 | tee -a $(ERRORLOG)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -134,12 +133,28 @@ build-images: #toil-docker
...
+#>>>
+# clean target: remove container, volumes, tempfiles
+# make clean-%
+
+#<<<
+clean-%:
+	echo "    started clean-$*"
+	source setup_hosts.sh
+	docker compose -f lib/candigv2/docker-compose.yml -f lib/$*/docker-compose.yml down || true
+	-docker volume rm `docker volume ls --filter name=$* -q`
```
OrdiNeu, line 154: This does create a sorta-hidden dependency on us to name our volumes correctly (i.e. each submodule should name their associated volume with $ somewhere in the title, and no two titles can overlap).

OrdiNeu: While trying to see if there was any corner cases, I found the following lines:
https://github.com/CanDIG/CanDIGv2/blob/develop/Makefile#L336-L337
I'm unsure if we use those two volumes anymore. Grafana is part of the old monitoring framework as far as I can see, and we no longer support Jupyter.

lib/vault/vault_setup.sh
```diff
@@ -31,11 +31,15 @@ do
   sleep 1
   docker ps --format "{{.Names}}" | grep vault_1
 done
-sleep 5
+sleep 10
```
OrdiNeu, line 34: Is there a better way to check if the keys can be grabbed other than this sleep? It might be worth it to just repeat the `docker exec` below until we get a valid response (or if Vault goes down as a failure case), as this might cause problems when trying to troubleshoot CanDIG on slower systems.

OrdiNeu (approved):
Looks good as-is, though I have a few thoughts below:

Makefile
daisieh: I think we've always just left the extra docker containers there as if we might bring them back, but if we want to take them out, we should do it as part of a separate PR.

daisieh: I can add some extra documentation to the container template to address the volume naming requirements. I have noticed that we don't actually document how to add an extra volume anyway.

lib/vault/vault_setup.sh
daisieh: I'm going to leave this alone for now, but will write myself a technical debt ticket to rethink vault setup separately.

PR #325, opened 2023-11-10
Only run minimal setup tests for htsget

daisieh:
As per a conversation in Slack, I think we can reduce the amount of time and complexity it takes to run integration tests on htsget if I only run the setup tests that we absolutely need.


daisieh:
this should really be happening via ingest integration tests anyway

PR #329, opened 2023-11-12
DIG-1272: Clinical ingest integration tests

daisieh:
Pick up changes from https://github.com/CanDIG/candigv2-ingest/pull/16 and rebuild. `make test-integration` should test the clinical ingest of single_ingest.json for both non-admin and admin users.


SonQBChau (approved):
LGTM

PR #332, opened 2023-11-13
Stable git history fixup

OrdiNeu:
Did a git merge from `develop` into `stable` (so that we can port back all the changes made for stable releases into develop and fix up the history), but unsure why it says there's 256 new commits. There's no files changed here so, hopefully it should be fine? After this I'll make a PR for stable->develop.


daisieh (approved):
Yeah, this seems fine to me?

PR #337, opened 2023-11-13
Stable merges back to develop

OrdiNeu:
This PR includes code changes from the prod server changes, which never got merged back into develop.


daisieh (approved):
Looks good to me. I cleaned out tyk and rebuilt it, just to make sure that the extra tyk config changes didn't break anything, and it seems fine.

daisieh:
The integration test failure here seems to be caused by a change to the submodule of Katsu that happened after the stable release. I think that it should resolve itself on the develop branch next time there's a PR.

PR #338, opened 2023-11-15
DIG-1272: htsget ingest tests

daisieh:
* Picking up changes from https://github.com/CanDIG/candigv2-ingest/pull/43
* Moving htsget tests to a later point and reducing the amount of htsget's own tests that need to be run for setup
* Added two tests, one for htsget ingest and one to check htsget's completeness data for a sample


OrdiNeu (approved):
Works and looks good to me

PR #351, opened 2023-11-27
Update test_integration.py

SonQBChau:
- update test integrations to work with new katsu uuid patch
- need to rebuild katsu
- most changes just syntax endpoint from `ingest/chemotherapies/` to `ingest/chemotherapy/`
- data in the body accept only an object instead of array of object


daisieh (changes requested):
Please update the katsu submodule so that the github action is testing the correct version.

daisieh (approved):
Everything passes now!

PR #358, opened 2023-12-18
DIG-1395: remove references for requirements-dev.txt

daisieh:
I grepped `grep -R "requirements-dev.txt" lib/` on the CanDIGv2 repo and got only two relevant hits, both of which were for commented-out lines. I think this is a clean deletion.


OrdiNeu (approved):
LGTM :+1: I found no results after this either

PR #371, opened 2024-01-02
DIG-1144: Update to python 3.12

daisieh:
All modules except for htsget should be able to move to 3.12 (and pip to 23.3.1).


OrdiNeu (approved):
Looks good to me, though we might need to upgrade in the future (and we should note when we should re-check if pysam's been fixed) :heavy_check_mark:

PR #374, opened 2024-01-04
Install and ingest doc mods.

DavidBrownlee:
Documentation modification resulting from my experience of installing CanDIGv2 on Ubuntu.  Mostly changes to the docker install, but a few corrections as well.

### DIFF
--- a/docs/ingest-and-test.md
+++ b/docs/ingest-and-test.md
@@ -17,7 +17,16 @@ pip install -r etc/venv/requirements
 make test-integration
 ```
 
+If the 4 federation tests fail, restart the federation container with:
+```bash
+make clean-federation
+make build-federation
+make compose-federation
+```
+
+
 If you want to run the tests manually, follow the instructions below.
+
 ## Initial tests
 
 Check that you can see the data portal in your browser at `http://candig.docker.internal:5080/`. If not, you may need to follow the instructions in the [Docker Deployment Guide](./install-docker.md)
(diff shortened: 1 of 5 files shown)


docs/ingest-and-test.md
```diff
@@ -6,11 +6,27 @@ These instructions will lead you through some basic functionality tests, ingesti
...
 make test-integration
 ```
 
+If the 4 federation tests fail, restart the federation container with:
+```bash
+make clean-federation
+make build-federation
+make compose-federation
+```
+
```
OrdiNeu, line 26:
```suggestion
```

etc/tests/tokenTest.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
OrdiNeu, line 1: While this is useful, I'm not sure that tests/ is the right folder for this. After a bit of discussion, Daisie and I found that there's no real spot for us to put helper scripts like this (that aren't part of the build process), so... I think here is fine until we reorganize it.

OrdiNeu (approved):

DavidBrownlee:
Thanks for the review.  I agree this isn't a great location for tokenTest.sh, but this is only a modification of a previous script (that was in the same location, but is now absent).  The new script pulls values from etc/secrets/... directly so it can be run from the command line without any copy & paste operations.

mshadbolt:
Is this ready to merge in? We have another PR open with doc changes so would be good to merge this in, then resolve any conflicts and merge the other one too

PR #376, opened 2024-01-04
Fixes #375 for query and candig-data-portal

### LINKED ISSUE #375: OPA server failing health check
Noticed that the container image `openpolicyagent/opa:latest-static` is basically an empty container with a static binary, so having a healtcheck based on `curl` will always fail.

Relevant output of docker ps:
```
818cc28fa2aa   openpolicyagent/opa:latest-static   "/opa run --server -…"   19 minutes ago   Up 19 minutes (unhealthy)   0.0.0.0:8181->8181/tcp, :::8181->8181/tcp                       candigv2_opa_1
```

EDIT: Same issue of missing the curl command for vault and minio.

query and candig-data-portal are also failing to get a healthy status, but they do have the curl command. It just fails.

stobbsm:
Health checks have been modified to check the internal container port instead of the host port.

Will look into fixing or finding a solution for the other containers mentioned


stobbsm:
I've created a quick go program to fill in for curl, which isn't in the remaining containers for a health check.
[gohttpcheck](https://github.com/stobbsm/gohttpcheck)
I will setup the containers in the other projects using a pull request to integrate this as a health check, avoiding the need for curl in the container itself. Since the healthcheck is just a get request using curl, all this application does is perform a head check to the given http address, to be used inside of the container environment.

Local testing has proved that it works.

mshadbolt:
Thanks for bringing this to our attention @stobbsm . We think we have implemented a fix to the unsuccessful health checks which we have now merged into develop (https://github.com/CanDIG/CanDIGv2/pull/401) and will make a part of our next stable release. Let us know if you are still having any issues!

PR #382, opened 2024-01-13
UHN prod hotfixes

daisieh:
These changes to https://github.com/CanDIG/CanDIGv2/releases/tag/v2.1.1 were made to UHN production to allow (almost) all of the integration tests to pass. The only one that does not pass (as of v2.1.1 + hotfixes) is within test_htsget, and we've since removed that test from test_integration.py anyway.


OrdiNeu (approved):
Works locally and the changes look good :heavy_check_mark: .

PR #391, opened 2024-01-19
Hotfix for UHN prod: clean should remove images

daisieh:
This is already inserted into UHN prod and it works.


SonQBChau:
worked for me!

SonQBChau (approved):

PR #393, opened 2024-01-19
DIG-1382, DIG-1384: query uses a service token with htsget

daisieh:
Query uses the new create/verify-service-token process in https://github.com/CanDIG/candigv2-authx/pull/14 to send an `X-Service-Token` header with its requests. Htsget, on receiving a beacon request with a valid `X-Service-Token` header for query, adds a `query-info` key into its response. For now, query just passes that on in its response as a `genomic_query_info` key, but that should be changed to something more useful.

There is a test added to test-integration, `test_query_info`, that checks this.


OrdiNeu (approved):
Seems good to me, thank you for setting this up so I could test it all at once :+1:

PR #403, opened 2024-02-01
Add integration test changes to be hg38, and also query integration tests

OrdiNeu:
# Description
This PR does two things and maybe should be split into two.
1. This switches around every instance of hg37 to hg38, which means that the frontend should be able to pull data properly. This includes the query fixes to the genomic search
2. This also adds the integration tests for query.

This requires the following PRs: 

- https://github.com/CanDIG/candigv2-ingest/pull/54
- https://github.com/CanDIG/query/pull/15
- https://github.com/CanDIG/query/pull/14
- https://github.com/CanDIG/htsget_app/pull/283

### DIFF
--- a/etc/tests/test_integration.py
+++ b/etc/tests/test_integration.py
@@ -647,6 +647,97 @@ def test_add_server():
     assert response.status_code == 200
 
 
+# Query Test: Get all donors
+def test_query_donors_all():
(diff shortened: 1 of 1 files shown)


daisieh (approved):
Looks good, tests well.

PR #421, opened 2024-02-07
Load ingest test data from the ingest repo in stack

mshadbolt:
This commit loads the `clinical_ingest.json` and `genomic_ingest.json` from the currently checked out repo in the stack, rather than the develop branch of ingest.

To test:

* Run `make test-integration`.

### DIFF
--- a/etc/tests/test_integration.py
+++ b/etc/tests/test_integration.py
@@ -282,8 +282,8 @@ def test_ingest_permissions():
     clean_up_program("SYNTHETIC-2")
     clean_up_program("TEST_2")
 
-    test_loc = "https://raw.githubusercontent.com/CanDIG/candigv2-ingest/develop/tests/clinical_ingest.json"
-    test_data = requests.get(test_loc).json()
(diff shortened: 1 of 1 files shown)


mshadbolt:
The test failures are expected because the version of clinical_ingest.json in the stack does not have date intervals but it refers to the katsu schema that does have date intervals, even though the katsu in the stack does not have date intervals. Therefore clinical ingest fails, and subsequently query can't query any clinical data.

I am not sure how or if we should fix this.

daisieh:
I don't think that's the case: the version of clinical ingest in here includes https://github.com/CanDIG/candigv2-ingest/pull/56, which fixed the clinical_ingest.json. The failure here is now this sporadic problem we've been seeing here: https://candig.atlassian.net/browse/DIG-1479

daisieh (approved):
LGTM

mshadbolt:
We will merge this anyway because the integration test failure is unrelated.

PR #427, opened 2024-02-08
Nightly build fixes?

OrdiNeu:
# Description:
Clean the bin between nightly builds (may have been related to the toil problems?) and also ensure that we're using the latest ingest script (clinical ETL now needs to be up-to-date)

### DIFF
--- a/nightly_build.sh
+++ b/nightly_build.sh
@@ -6,7 +6,6 @@ PostToSlack () {
     curl -X POST -H 'Content-type: application/json' --data "{\"text\":\"$SAFE_TEXT\"}" $HOOK_URL
 }
 
-
 # Make sure all of our necessary configuration works
 source nightly_env.sh
 if [ -z "$HOOK_URL" ] || [ -z "$BOT_TOKEN" ]; then
@@ -16,6 +15,7 @@ fi
 
 # Clean all
 make clean-all
+make clean-bin
 docker system prune -af
 
 # Double check that the .env file works?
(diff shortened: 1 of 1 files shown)


nightly_build.sh
```diff
@@ -81,8 +81,11 @@ source env.sh
 cd $INGEST_PATH
 export CLINICAL_DATA_LOCATION=$INGEST_PATH/tests/clinical_ingest.json
 # should be pip install -r requirements.txt, but that didn't seem to work last I checked -- dependency errors?
+# `tox 4.12.1 requires colorama>=0.4.6, but you have colorama 0.4.4 which is incompatible.`
+git pull
 pip install dateparser
 pip install openapi_spec_validator
+pip install clinical_etl@git+https://github.com/CanDIG/clinical_ETL_code.git@v2.0.0
```
daisieh, line 88: Is there a reason you're not doing `pip install -r etc/venv/requirements.txt` here?

daisieh (changes requested):
one comment

nightly_build.sh
OrdiNeu: Trying to run that crashes with `tox 4.12.1 requires colorama>=0.4.6, but you have colorama 0.4.4 which is incompatible.`. There might be a better fix somehow, but I'm unsure how to resolve

daisieh: weird...if you look at pip list now, it seems like it has colorama 0.4.6 installed.

OrdiNeu:
Closed because these fixes don't appear to be working on dev anymore

PR #436, opened 2024-02-13
DIG-1459: Async indexing in htsget

daisieh:
Adds a test to test-integration for indexing.

Also sets INDEXING_PATH in htsget's docker-compose file and updates submodules.


OrdiNeu (approved):
LGTM, just needs a comment that it's re-checking something that must already be indexed by this point :heavy_check_mark:

PR #468, opened 2024-02-28
Cleaning up test data should clean up htsget cohorts too

daisieh:
I forgot that I never implemented this!


OrdiNeu (commented):
LGTM, though there's a remaining cohort after running integration tests -- test-htsget. Should that get cleaned up as well?

daisieh:
htsget tests need to be explicitly cleaned up...I have added that now.

OrdiNeu (approved):
LGTM :heavy_check_mark:

PR #478, opened 2024-03-02
DIG-1511: htsget tests need to be more careful about user

daisieh:
After fixing https://github.com/CanDIG/htsget_app/pull/295, I realized how many integration tests were dependent on the site admin user being able to see everything in htsget. I've updated the tests so that they work for the specific user that should be allowed to see the data.


OrdiNeu (approved):
LGTM :heavy_check_mark:

PR #487, opened 2024-03-07
hotfix: set KEEP_TEST_DATA to false by default

daisieh:
Switch the setting in your .env back and forth to true/false. If it's true, you'll get:
```
$ make test-integration
python ./settings.py
source ./env.sh; pytest ./etc/tests -k 'not test_clean_up'
```

If it's false, you'll get:
```
$ make test-integration
python ./settings.py
source ./env.sh; pytest ./etc/tests 
```


OrdiNeu (approved):
Works for me :heavy_check_mark:

mshadbolt (approved):
Looks good, Thank you!

PR #488, opened 2024-03-07
DIG-1520: env changes for Opa PR #51

daisieh:
Creates the CANDIG_USER_KEY and passes it to the Opa microservice. This key allows the site to adjust the unique identifier used in their IDP's jwts, in case they don't use email (or name it differently).

This picks up https://github.com/CanDIG/candig-opa/pull/51.


kcranston:
We should also remove the admin setup in Keycloak, if we are going to use OPA roles for admin.

kcranston (approved):

PR #490, opened 2024-03-07
ingest-and-test.md documentation changes only.

DavidBrownlee:
Much more clear now.


mshadbolt (approved):
Made a little edit to include how to use the KEEP_TEST_DATA to keep the data ingested during integration tests.

Thanks for improving our docs!

PR #494, opened 2024-03-14
Fix make clean-*

mshadbolt:
Found that make clean was not working for individual services and discovered the tag needed to be incorporated to delete properly. Maybe a docker versioning thing? Would be good to have others test whether this fix works for them too.


OrdiNeu (approved):
I was able to remove every image one by one with this change :+1:

daisieh (approved):
Looks good to me, and works on my new Docker build. It doesn't manage to remove images with no tag, like `candig/candig-ingest:<none>`, but nothing seems to work for that. You have to do a manual `docker image prune` to deal with those, I think.

PR #497, opened 2024-03-14
change to quay.io keycloak

mshadbolt:
since the jboss image was removed we are changing to the quay.io keycloak temporarily until we upgrade to a later version.


CourtneyGosselin (approved):
LGTM!

PR #504, opened 2024-03-22
Fix the CIDR 16-bit subnet mask replacement with a general-purpose one

OrdiNeu:
Vault service store creation fails when the docker container's IP address's subnet mask is not 16 bits wide.

Example:
```
create an approle for opa
Error writing data to auth/approle/role/opa: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/auth/approle/role/opa
Code: 500. Errors:

* 1 error occurred:
	* failed to validate CIDR blocks: invalid CIDR address: 192.168.48.13/2032
```
Here, a substitution for /20 failed, and it ended up as a 2032 bit sized block.


daisieh (approved):
Looks good, thanks for catching these!

PR #507, opened 2024-03-23
DIG-1517, DIG-1518: Opa/authx changes for role-based and program-based authorizations

daisieh:
Picks up changes from https://github.com/CanDIG/candigv2-authx/pull/19, https://github.com/CanDIG/candig-opa/pull/52, https://github.com/CanDIG/candigv2-ingest/pull/80, and https://github.com/CanDIG/htsget_app/pull/297.

Integration tests now test adding/getting/deleting program authorizations via candig-ingest (and thus tests the opa and authx connections as well). Htsget needed to have some of its methods and tests tweaked to match as well.

### DIFF
--- a/etc/tests/test_integration.py
+++ b/etc/tests/test_integration.py
@@ -119,7 +119,7 @@ def test_opa_datasets(user, dataset):
 
 
 ## Can we add a dataset to one of the users?
-def test_add_opa_dataset():
+def test_add_remove_opa_dataset():
     token = get_token(
         username=ENV["CANDIG_SITE_ADMIN_USER"],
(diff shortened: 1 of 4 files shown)


OrdiNeu (commented):
Quick q: - there's a bunch of changes to how we authorize specific endpoints in the HTSGet PR; won't these authz endpoint changes need to be applied to Katsu as well?

daisieh:
I don't think so. The htsget changes were mostly tweaks to take advantage of the new is_action_allowed method and a few similar refactors.

OrdiNeu (approved):

PR #514, opened 2024-03-27
DIG-1549: Ingest uses private urls if running within a container

daisieh:
Ingest should use the private versions of container URLs if possible: we will pass in private URLs via docker-compose.


OrdiNeu (approved):
LGTM :heavy_check_mark:

PR #517, opened 2024-03-27
Extend timeout for all queries to be 60s

OrdiNeu:
[Jira ticket](https://candig.atlassian.net/browse/DIG-1551)

This was only tested a little bit, by inserting random time.sleep() into a few random places in Katsu.


daisieh (approved):
Yep, makes sense

PR #522, opened 2024-03-28
Hotfix: Fix check for errors in test_ingest_htsget

daisieh:
We decided to always leave the "errors" key in the dict, even if it's empty.


mshadbolt (approved):
looks good and tests passing

PR #526, opened 2024-04-05
Allow usage of both KEEP_TEST_DATA and ARGS in the same command

OrdiNeu:
Small convenience command that fixes an issue Daisie found out a while ago and that I just ran into this morning; this prevents KEEP_TEST_DATA from overriding the ARGS argument.

# To Test:
- `make test-integration KEEP_TEST_DATA=true ARGS="-v -k 'test_tyk'"` now runs just one test with full verbosity


SonQBChau (approved):
LGTM! tested and it worked.

PR #527, opened 2024-04-08
DIG-1532, DIG-1443: separate site admin user from other test users

daisieh:
I couldn't quite figure out how to separate these two tickets.

* Cleaned up the deprecated OPA_SITE_ADMIN_KEY from docker-compose files, settings, etc.
* Made the test usernames consistently email addresses (in our test case; this could be changed on different systems by changing CANDIG_USER_KEY in the env file.)
* Created a new script, site_admin_token.py, that can be used for things like test-integration. It looks for either a stored site admin username/password or asks for those, then saves off a refresh token that can be used with the bumped authx library to repeatedly get an access token as needed.
* (this script can replace tokenTest.sh, as it handles more cases than the default site admin user and password)
* Federation setup (initialize.py) and opa_setup.py use this new script for getting a site admin token. 
* Created a separate keycloak test user, DEFAULT_SITE_ADMIN_USER=site_admin@test.ca, that will be interpolated as the default test user and is the one that should be replaced on production systems.

The submodules opa and htsget have changes as well, but they're minor and related to cleaning up how site admin and other users get passed in.
https://github.com/CanDIG/candig-opa/pull/53
https://github.com/CanDIG/htsget_app/pull/298


OrdiNeu:
Does this change prevent you from logging in as user2 to the frontend?

daisieh:
If user2 is still set as the default site admin, you should still be able to log in.

etc/tests/test_integration.py
```diff
@@ -591,10 +565,7 @@ def test_verify_htsget(object_id, file_name, file_type, user):
             password=ENV["CANDIG_NOT_ADMIN_PASSWORD"],
         )
     elif user == "user2":
-        token = get_token(
-            username=ENV["CANDIG_SITE_ADMIN_USER"],
-            password=ENV["CANDIG_SITE_ADMIN_PASSWORD"],
-        )
+        token = get_site_admin_token()
```
OrdiNeu: This never seems to actually get called -- verify_samples only ever uses user1

OrdiNeu (approved):
LGTM :heavy_check_mark:

etc/tests/test_integration.py
daisieh: fixed

PR #534, opened 2024-04-11
hotfix: now that awscli requirement is loosened, should be able to install requirements

daisieh:
requirements in ingest were updated in https://github.com/CanDIG/candigv2-ingest/pull/86


OrdiNeu (approved):
LGTM :heavy_check_mark:

PR #539, opened 2024-04-16
DIG-1243: Set up and configure Redis independent from tyk

SonQBChau:
## What's new
- Redis now has its own module
- Rename volume `tyk-redis-data` to `redis-data`
- Redis now requires a password to login
- Add Redis secret to tyk compose
- Add Redis secret to katsu compose

### How to test
- Rebuild the stack
- Run test-integration


daisieh (approved):
Looks good and seems to behave as before!

PR #543, opened 2024-04-22
DIG-1601: Remove version line from docker-compose files

OrdiNeu:
[Jira ticket](https://candig.atlassian.net/browse/DIG-1601)

This removes the version number from the top of our docker-compose.yml files. This should reduce the number of warnings we get from the full build, without changing anything else.

NB: There's a few other changes in here that my vim auto-added -- it looks like they're just formatting fixes (newline at end of file, tab correction)


mshadbolt (approved):
Wow so fast, LGTM!

PR #545, opened 2024-04-22
DIG-1588: set minimum threshold var

SonQBChau:
## What's new
- Add `AGGREGATE_COUNT_THRESHOLD` to example.env

### How to test
- `cp -i etc/env/example.env .env`
- Edit .env if needed
- Rebuilt the stack and run test-integration


mshadbolt (approved):
LGTM

PR #550, opened 2024-04-26
DIG-1496, DIG-1502: pending users, user authz

daisieh:
Fixed some misc small things, but mostly this one sweeps up changes in submodules and adds tests for new Opa/authx user authz.


daisieh:
Francis found a weird glitch that never surfaced before until I put the `is_local_token` clause into Opa in https://github.com/CanDIG/candig-opa/commit/b6cd636097484ab4a471254f8f3514bc586d7d48. It seems like idp.rego doesn't like the `id_token` that the front end passes in via Tyk...we've known that was wrong and should be fixed anyway, see https://candig.atlassian.net/browse/DIG-1541, so I added that fix into this PR.

OrdiNeu (approved):
After that last change, everything looks good to me :heavy_check_mark:

PR #562, opened 2024-05-01
DIG-1603: Confirm the stack can be taken down and rebuilt without deleting postgres

OrdiNeu:
# Description
This adds a new Makefile target: `make rebuild-without-postgres`, which rebuilds the stack without touching the Postgres container. This target is intended for production instances that want to redeploy the stack without taking down any of the ingested data (up to a point; as long as HTSGet/Katsu don't require database changes this should work).

# To test
1. Build the stack
2. Ingest data
3. Run `make rebuild-without-postgres`, optionally changing something in the data portal or something to confirm that those services will be rebuilt
4. Login again. You will need the new credentials, but all of the ingested data will still be there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -544,3 +544,28 @@ stop-all:
 start-all:
 	CONTAINERS="$(shell docker ps -a --format '{{.Names}}' | grep candigv2)"; for CONTAINER in $$CONTAINERS; do docker start $$CONTAINER; done
 
+#>>>
+# rebuild the entire stack without touching the postgres container
+
+#<<<
+.PHONY: rebuild-without-postgres
+rebuild-without-postgres:
+	# Back up postgres-related variables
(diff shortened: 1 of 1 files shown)


mshadbolt (approved):
Tested this on my local stack and everything worked well. 

I assume that someone would need to redo the program authorizations for users?

Maybe we can have documentation that would tell a site admin how to back up the program authorizations and re post them after the rebuild?

OrdiNeu:
@mshadbolt 
> I assume that someone would need to redo the program authorizations for users?

Yes. That would require us to keep Vault/opa alive which... might be possible? Unsure at the moment.

> Maybe we can have documentation that would tell a site admin how to back up the program authorizations and re post them after the rebuild?

I think we'd want a ticket for that -- I'm not sure how to go about that

PR #579, opened 2024-05-09
Hotfixes for stable

daisieh:
These are changes that mostly affect UHN prod.
* Because it has a weird thing where it doesn't always know about its own hostname, it often needs to refer to itself as 127.0.0.1 directly. I split out the references to CANDIG_DOMAIN into CANDIG_DOMAIN and CANDIG_INTERNAL_DOMAIN, so that integration tests run from the server itself will be able to use CANDIG_INTERNAL_DOMAIN if needed.
* I added test_install_public_object to the htsget integration tests, so that if minio isn't connecting correctly (as on UHN prod), we are still testing some sort of s3 object.
* I had to add an internal URL for query so that test_query_discovery would be able to use the correct URL.
* I also needed to make sure that VAULT_SERVICE_PUBLIC_URL is accessible as an env var so that vault_setup works correctly.

These are already running on prod, and github testing will test it overall, but it wouldn't hurt to do a local test as well. Install and run integration tests.


SonQBChau (approved):
Passed integration tests 👍

PR #585, opened 2024-05-10
Removing minio from the stack

daisieh:
We are probably best off not loading up the Minio container if we don't have to: now that it's not used for htsget, I don't think it's used for anything else.


SonQBChau (approved):
Tests passed. LGTM!

PR #589, opened 2024-05-13
DIG-1651: minimizing secrets

daisieh:
* Moved user names to example.env instead of having them in secrets
  * keycloak-test-site-admin
  * keycloak-test-user
  * keycloak-test-user2
* DEFAULT_ADMIN_USER can be set as "admin" in example.env instead of being hard-coded throughout and kept as a secret
  * keycloak-admin-user
  * metadata-db-user
* Removed many unused secrets:
  * tyk-node-secret-key
  * metadata-app-secret
  * ssl stuff
* keycloak-client-local_candig-id-64 isn't a secret; it can be generated when needed in tyk_preflight.sh
* keycloak-public-key isn't a secret; it can be generated when needed in tyk_preflight.sh
* katsu-secret-key can be generated and saved within the Katsu container (https://github.com/CanDIG/katsu/pull/206)

To test: run integration tests; things should still work.


OrdiNeu (approved):
Looks good to me :heavy_check_mark: NB that this involves a .env change when merged in

PR #592, opened 2024-05-15
DIG-1653: S3 token should be stored as a secret inside ingest's vault store

daisieh:
Instead of using the VAULT_S3_TOKEN directly to access a special Vault kv store, we can use candig-ingest's Vault service store. Htsget is the only other service that needs access to the store, so we allow it in vault_setup.sh (and in htsget_setup.sh.)

I modified the test_vault test to be test_aws_credentials to test the changes.


etc/tests/test_integration.py
```diff
@@ -308,46 +308,40 @@ def test_add_remove_site_admin():
 
 
 ## Vault tests: can we add an aws access key and retrieve it?
-def test_vault():
+def test_aws_credentials():
```
OrdiNeu:
```suggestion
def test_s3_credentials():
```

OrdiNeu (approved):

PR #594, opened 2024-05-17
Hotfixes for candig.sd4h.ca.  Many changes to example.env

DavidBrownlee:
Hotfixes required to get test.candig.sd4h.ca and candig.sd4h.ca (prod) working with stable V4.1.0.

Mostly fixes to the example.env file.

No submodules updated.


etc/env/example.env
```diff
@@ -8,7 +8,7 @@ CANDIG_AUTH_MODULES=keycloak vault tyk opa
 
 # options are [<ip_addr>, <url>, host.docker.internal, candig.docker.internal]
 CANDIG_DOMAIN=candig.docker.internal
-CANDIG_INTERNAL_DOMAIN=candig.docker.internal
+CANDIG_INTERNAL_DOMAIN=127.0.0.1
```
daisieh, line 11:
```suggestion
CANDIG_INTERNAL_DOMAIN=${CANDIG_DOMAIN}
```

daisieh: I made this change in a different PR, but we can make this change here as well.

etc/env/example.env
```diff
@@ -85,7 +85,7 @@ MINIO_SELF_CERT=0
 HTSGET_VERSION=v4.1.0
 HTSGET_PRIVATE_URL=http://htsget:3000
 HTSGET_PUBLIC_URL=${TYK_LOGIN_TARGET_URL}/${TYK_HTSGET_API_LISTEN_PATH}
-HTSGET_PORT=3333
+HTSGET_PORT=3000
```
daisieh, line 88: Why are you proposing to change these ports everywhere, instead of just in the C3G .env file?

etc/env/example.env
```diff
@@ -130,9 +130,8 @@ TOIL_WORKER_PORT=5051
 FEDERATION_VERSION=v2.1.1
 FEDERATION_IP=0.0.0.0
 FEDERATION_PORT=4232
-FEDERATION_SERVICE_URL=http://${CANDIG_INTERNAL_DOMAIN}:${FEDERATION_PORT}
+FEDERATION_SERVICE_URL=http://federation:${FEDERATION_PORT}
```
daisieh, line 133: This is supposed to be different than FEDERATION_PRIVATE_URL: in a context like federation_setup.sh, which is not called inside the container, the URL needs to be an internal one, not the container URL.

etc/env/example.env
```diff
@@ -181,9 +180,9 @@ KEYCLOAK_HOST=0.0.0.0
 KEYCLOAK_PUBLIC_PROTO=http
 KEYCLOAK_PRIVATE_PROTO=http
 KEYCLOAK_ENABLE_PROXY=false
-KEYCLOAK_PUBLIC_URL=${KEYCLOAK_PUBLIC_PROTO}://${CANDIG_AUTH_DOMAIN}:${KEYCLOAK_PORT}
+KEYCLOAK_PUBLIC_URL=${KEYCLOAK_PUBLIC_PROTO}://${CANDIG_AUTH_DOMAIN}
 KEYCLOAK_PUBLIC_URL_PROD=${KEYCLOAK_PUBLIC_PROTO}://${CANDIG_AUTH_DOMAIN}
-KEYCLOAK_PRIVATE_URL=${KEYCLOAK_PRIVATE_PROTO}://${CANDIG_AUTH_DOMAIN}:${KEYCLOAK_CONTAINER_PORT}
+KEYCLOAK_PRIVATE_URL=${KEYCLOAK_PRIVATE_PROTO}://keycloak:${KEYCLOAK_CONTAINER_PORT}
```
daisieh, line 185: We should try this value out on other servers before making this switch for sure, but I think this is probably ok?

etc/env/example.env
```diff
@@ -276,7 +275,7 @@ TYK_INGEST_API_LISTEN_PATH=ingest
 ## api - query
 TYK_QUERY_API_ID=102
 TYK_QUERY_API_SLUG=query
-TYK_QUERY_API_TARGET=${QUERY_PRIVATE_URL}
+TYK_QUERY_API_TARGET=http://query:${QUERY_PORT}
```
daisieh, line 278: Why did you change this? this one is the same value as before?

etc/env/example.env
```diff
@@ -264,7 +263,7 @@ TYK_VAULT_API_LISTEN_PATH=vault
 ## api - federation
 TYK_FEDERATION_API_ID=91
 TYK_FEDERATION_API_SLUG=federation
-TYK_FEDERATION_API_TARGET=${FEDERATION_PRIVATE_URL}
+TYK_FEDERATION_API_TARGET=${FEDERATION_SERVICE_URL}
```
daisieh, line 266: This should be the PRIVATE_URL, as in the other services.

daisieh (changes requested):
There are many changes to the example.env that I'd like to have explanations for, before accepting these changes as being global for all of our servers.

etc/env/example.env
DavidBrownlee: CANDIG_INTERNAL_DOMAIN=${CANDIG_DOMAIN} will not work since ${CANDIG_DOMAIN} is (in a server environment) set to the external URL.

UHN prod currently runs:
CANDIG_INTERNAL_DOMAIN=127.0.0.1

daisieh: Exactly...it is set specifically to a value for UHN prod, but should by default be the same as CANDIG_DOMAIN unless otherwise specified.

DavidBrownlee:
Hi Daisie,
Thank you for your thoughtful consideration of this PR.  I'm going to run a few more tests locally and I will get back to this PR.

daisieh:
Please only keep the absolutely necessary changes for all servers in this PR...if something is specific to C3G, it should be noted in the production confluence page, but not reflected in example.env.

DavidBrownlee:
Given the networking differences between personal development and deployed server environments, it seems any harmonized .env (or example.env) file is likely doomed at this point.

PR withdrawn.

PR #595, opened 2024-05-21
Fix the usernames being pulled for the nightly build

OrdiNeu:
# Description
The past few days the nightly build bot has been unable to determine the username for any of the default users -- this fixes that.


mshadbolt (approved):
Thanks for the fix!

PR #597, opened 2024-05-21
DIG-1664: Do not log service-info (prevent health check flooding)

OrdiNeu:
This aggregates PRs sent to multiple microservices that silences the service-info endpoint from showing up in the logs, which was causing flooding. This does not interfere with the operation of the health check itself. In HTSGet's and Katsu's case, the healthcheck was also fixed.

# Affected Microservices
- [Federation](https://github.com/CanDIG/federation_service/pull/57)
- [Query](https://github.com/CanDIG/candigv2-query/pull/38)
- [HTSGet](https://github.com/CanDIG/htsget_app/pull/310) + health check fix
- [Katsu](https://github.com/CanDIG/katsu/pull/207/files) + health check fix
- Tyk (health check fixed)

### DIFF
--- a/lib/htsget/docker-compose.yml
+++ b/lib/htsget/docker-compose.yml
@@ -38,7 +38,7 @@ services:
     extra_hosts:
       - "${CANDIG_DOMAIN}:${LOCAL_IP_ADDR}"
     healthcheck:
-      test: ["CMD", "curl", "--fail", "http://${CANDIG_DOMAIN}:3333/ga4gh/drs/v1/service-info"]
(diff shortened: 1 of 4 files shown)


daisieh (approved):
Looks good and works for me.

daisieh:
Even though Opa doesn't use uwsgi (or anything) so it doesn't log, there is no reason that it shouldn't match the others, so I fixed it.

OrdiNeu:
Force push to fixup merge conflict

PR #598, opened 2024-05-22
Fixup site admin & user1 not being able to access TEST_2

OrdiNeu:
# Description
TEST_2 was never ingested into OPA, which caused the Site Admin to be unable to view any patients in TEST_2. This fixes that up by giving those permissions to user1.


daisieh (approved):
Looks good, thanks!

mshadbolt (approved):
works for me too!

PR #620, opened 2024-05-30
DIG-979: Move federation config to Vault service store

daisieh:
Using the new version of federation in https://github.com/CanDIG/federation_service/pull/60, we can now initialize our services and servers with the values stored in federation's vault service store. This should still pass integration tests.

### DIFF
--- a/lib/federation/initialize.py
+++ b/lib/federation/initialize.py
@@ -40,13 +40,6 @@ def get_default_server():
 def main():
     token = get_site_admin_token()
 
-    server = {
-        "server": json.loads(get_env_value("FEDERATION_SELF_SERVER").replace('\'', '"')),
-        "authentication": {
-            "issuer": get_env_value("KEYCLOAK_REALM_URL"),
-            "token": token
-        }
-    }
     headers = {}
     headers["Authorization"] = f"Bearer {token}"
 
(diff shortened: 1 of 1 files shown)


lib/federation/initialize.py
```diff
@@ -37,31 +50,34 @@ def main():
...
 
-    print("Adding servers to federation...")
+    print("Register existing servers")
     url = f"{get_env_value('FEDERATION_PUBLIC_URL')}/v1/servers"
+    response = requests.request("POST", url, headers=headers, params={"register": True})
+    if response.status_code != 200:
+        print(f"POST response: {response.status_code} {response.text}")
+
+    print("Making sure our server is in federation...")
+    server = get_default_server()
```
OrdiNeu, line 53: Best to remove the server = {} json in lines 43-49

lib/katsu/katsu_service
OrdiNeu, line 1: What's the Katsu/OPA changes in this PR?

OrdiNeu (approved):
Works. Built federation, then cleaned it and edited the setup to remove adding servers/services. Rebuilt it and the old servers/services were still there 👍

lib/federation/initialize.py
daisieh: oh yes, good point.

PR #623, opened 2024-06-03
DIG-1637: make target that keeps containers with data

daisieh:
The modules `keycloak vault redis postgres` are now considered `CANDIG_DATA_MODULES`. These are containers that tend to contain data that we want to retain in between stable releases; they are not based on our own code and only really need to be rebuilt when we do a major update of their underlying software.

I moved their secrets to `tmp/module/` so that we can more clearly keep or regenerate those secrets as needed; the secrets will still be created by `make docker-secrets`, but when the module is built, they will be moved into the appropriate tmp directory and deleted from the tmp/secrets folder. The modules will delete these tmp/module directories only when cleaned.

If Keycloak is recomposed, we will end up with a new client-secret that needs to be propagated to Opa's vault secret for its IdP providers: that is picked up by candigv2-authx@v2.4.3.

We can now simplify (and rename to `rebuild-keep-data`) the `rebuild-without-postgres` make target: secrets don't need to be touched directly, because they are handled by individual modules. We can just filter out the CANDIG_DATA_MODULES and rebuild all the other modules.


OrdiNeu:
Is the intent to just run `make rebuild-keep-data` from a running instance? If so, building is fine but after rebuilding OPA it fails out with the following:
```
if [ -f lib/opa/opa_setup.sh ]; then \                                                                                                                                                                      
source lib/opa/opa_setup.sh 2>&1 | tee -a tmp/error.txt; \                                            
fi                                                                                                                                                                                                          
Traceback (most recent call last):                                                                                                                                                                          
  File "/home/fnguyen/Documents/CanDIGv2/settings.py", line 83, in <module>                                                                                                                                 
    main()                                                                                                                                                                                                  
  File "/home/fnguyen/Documents/CanDIGv2/settings.py", line 76, in main                                                                                                                                     
    vars = get_env()                                                                                  
           ^^^^^^^^^                                                                                                                                                                                        
  File "/home/fnguyen/Documents/CanDIGv2/settings.py", line 56, in get_env                            
    with open(f"tmp/secrets/keycloak-client-{vars['CANDIG_CLIENT_ID']}-secret") as f:                 
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                       
FileNotFoundError: [Errno 2] No such file or directory: 'tmp/secrets/keycloak-client-local_candig-secret'                                                                                                   
>> waiting for opa_runner to start                                                                    
candigv2_opa-runner_1                                                                                                                                                                                       
candigv2_vault_1                                                                                                                                                                                            
>> setting up opa store                                                                                                                                                                                     
create a policy for opa's access to the approle role and secret                                                                                                                                             
Success! Uploaded policy: opa
create an approle for opa
Success! Data written to: auth/approle/role/opa
>> setting up opa store policy
Success! Uploaded policy: opa
save the role id to secrets
create a kv store for opa
Error enabling: Error making API request.

URL: POST http://127.0.0.1:8200/v1/sys/mounts/opa
Code: 400. Errors:

* path is already in use at opa/
Error response from daemon: Could not find the file /home/candig in container candigv2_opa_1
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/fnguyen/Documents/CanDIGv2/site_admin_token.py", line 5, in <module>
    ENV = get_env()
          ^^^^^^^^^
  File "/home/fnguyen/Documents/CanDIGv2/settings.py", line 56, in get_env
    with open(f"tmp/secrets/keycloak-client-{vars['CANDIG_CLIENT_ID']}-secret") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'tmp/secrets/keycloak-client-local_candig-secret'
candigv2_opa-runner_1
candigv2_opa_1

```

daisieh:
I think there was a merge conflict in settings.py with another change that got merged into dev today...try it again?

OrdiNeu:
Slightly different issue during rebuilding:
```
create a kv store for opa
Error enabling: Error making API request.

URL: POST http://127.0.0.1:8200/v1/sys/mounts/opa
Code: 400. Errors:

* path is already in use at opa/
Error response from daemon: Could not find the file /home/candig in container candigv2_opa_1
Traceback (most recent call last):
  File "/home/fnguyen/Documents/CanDIGv2/site_admin_token.py", line 33, in get_site_admin_token
    client_secret=ENV["CANDIG_CLIENT_SECRET"],
                  ~~~^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'CANDIG_CLIENT_SECRET'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 2, in <module>
  File "/home/fnguyen/Documents/CanDIGv2/site_admin_token.py", line 47, in get_site_admin_token
    raise authx.auth.CandigAuthError(f"Error obtaining response from keycloak server: {e}")
authx.auth.CandigAuthError: Error obtaining response from keycloak server: 'CANDIG_CLIENT_SECRET'
candigv2_opa-runner_1
candigv2_opa_1
```
And then federation still fails to build.

OrdiNeu:
It looks like the keycloak secrets might need the same treatment that vault/redis have in this PR.

daisieh:
Oh, I guess that is true: if we're not rebuilding keycloak, then we do need to keep its secrets. I'll fix that.

daisieh:
Okay I made a bunch of changes and I think it's good now?

OrdiNeu (approved):
Works well :+1:

PR #624, opened 2024-06-03
Add DEFAULT_ADMIN_USER to settings.py

daisieh:
Update a few loose bits in settings.py and make sure that compose-% gets the latest versions.


CourtneyGosselin (approved):
LGTM Build and integration tests pass works with added work here https://github.com/CanDIG/CanDIGv2/pull/627

PR #627, opened 2024-06-04
sed added to work on WSL

CourtneyGosselin:
## Description 
Secrets were not building correctly narrowed down to the format returned from `CLIENT_SECRET=$(docker exec "$KEYCLOAK_CONTAINER_ID" /opt/keycloak/bin/kcadm.sh get clients/"$CLIENT_ID"/client-secret -r "$KEYCLOAK_REALM" |  jq -r '.value')` wasn't what jq expected added sed.

@SonQBChau feel free to make changes


daisieh:
This should work just fine: I don't know why the command doesn't only output json on wsl, but if we can sed out just the json, we should be fine.

daisieh:
To be clear: on my local version, the output of `$(KCADM get clients/"$CLIENT_ID"/client-secret -r "$KEYCLOAK_REALM")` is just json for me. On WSL, the output includes a bit of prefix: 
```
get clients/<client_id>/client-secret -r candig { "type" : "secret", "value" : "value" }
```

SonQBChau (approved):
Did not cause any problem on my machine and it passed github action

PR #635, opened 2024-06-13
Miscellaneous fixes, fix DIG-1673

daisieh:
Little tweaks that I ran into while using the new test suite for my development.

(Fixes DIG-1673) I also realized that the `docker restart` in htsget_setup.sh was cutting off the loading of the ncbirefseq table, so I removed it. I don't think it's necessary; I just had it in the template.


mshadbolt (approved):
LGTM and passes integration tests on my local

PR #540, opened 2024-04-17
DIG-1529: Integration tests use new synthetic data

mshadbolt:
Should be merged after https://github.com/CanDIG/CanDIGv2/pull/613

- Adds to `ingest_setup.sh` to generate the small_dataset ingest jsons and save them into the `lib/candig-ingest/candigv2-ingest` repo
- Uses the new synthetic data to ingest as part of integration tests
- Modifies several integration tests to work with the new data
- Changes  a few of the tokens from site admin to specific users 
- Adds a new target to makefile to run an individual test and print output whether passing or not `test-integration-%` where % is the name of the test
- Propose deleting running the htsget unit tests
- Propose deleting the add samples to htsget test `test_htsget_add_sample_to_dataset()`

PR #642, opened 2024-06-14
Remove GraphQL, which was causing the frontpage to no longer load

OrdiNeu:
# Description
#540 introduced a regression where the frontpage would no longer load, and the redirect to the Keycloak login was similarly failing. This was caused by the GraphQL service, which is no longer in use but still having an entry in Tyk, getting pointed to the root URL and eating away both the redirect and the Data Portal. This PR removes the endpoint entirely from Tyk, and should allow the frontpage and login redirect to function as expected, without losing any of the other changes from #540 .


mshadbolt (approved):
portal works and tests pass, (except ingest one which keeps timing out) LGTM

PR #643, opened 2024-06-14
Update actions versions

mshadbolt:
Some of our actions are on older versions that may be deprecated soon, updating to avoid that and get rid of the warning messages when running actions

To test:
* compare to previous build https://github.com/CanDIG/CanDIGv2/actions/runs/9519738192 and note that there are no longer warnings at the bottom


SonQBChau (approved):
Tested and no more warning

PR #644, opened 2024-06-14
set keep_test_data in github action

daisieh:


mshadbolt (approved):
github build passing, LGTM

PR #645, opened 2024-06-14
Modify ingest test and build nightly with no cache

mshadbolt:
* Changed the `test_ingest_admin_katsu` test to check success by checking what is ingested into katsu instead of waiting for a response, which tends to timeout
* Added some print statements to help with debugging
* Commented out clinical ingest response checks since it often times out
* added no-cache to the BUILD_OPTS in nightly build

### DIFF
--- a/etc/tests/test_integration.py
+++ b/etc/tests/test_integration.py
@@ -388,14 +388,15 @@ def clean_up_program_htsget(program_id):
 
 
 def test_ingest_not_admin_katsu():
-    katsu_response = requests.get(f"{ENV['CANDIG_ENV']['KATSU_INGEST_URL']}/v2/discovery/programs/").json()
-    katsu_programs = [x['program_id'] for x in katsu_response]
-    if 'SYNTHETIC-1' in katsu_programs:
-        print("cleaning up 'SYNTHETIC-1'")
-        clean_up_program("SYNTHETIC-1")
-    if 'SYNTHETIC-2' in katsu_programs:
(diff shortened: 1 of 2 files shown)


etc/tests/test_integration.py
```diff
@@ -385,8 +388,14 @@ def clean_up_program_htsget(program_id):
 
 
 def test_ingest_not_admin_katsu():
-    clean_up_program("SYNTHETIC-1")
-    clean_up_program("SYNTHETIC-2")
+    katsu_response = requests.get(f"{ENV['CANDIG_ENV']['KATSU_INGEST_URL']}/v2/discovery/programs/").json()
```
daisieh: do you need to check to make sure that this returns a json-able response? what if it's not status_code == 200?

etc/tests/test_integration.py
```diff
@@ -406,8 +415,14 @@ def test_ingest_not_admin_katsu():
 
 
 def test_ingest_admin_katsu():
-    clean_up_program("SYNTHETIC-1")
-    clean_up_program("SYNTHETIC-2")
+    katsu_response = requests.get(f"{ENV['CANDIG_ENV']['KATSU_INGEST_URL']}/v2/discovery/programs/").json()
```
daisieh: Same here: check for status_code == 200?

daisieh (changes requested):
Make sure there are json-able responses before calling response.json()

daisieh (approved):
LGTM!

PR #657, opened 2024-06-19
DIG-1690: Cannot logout on dev bugfix

OrdiNeu:
# Description

This fixes the bug where you can't logout.

# To test
1. Sign into the data portal
2. Sign out
3. You should be redirected to the signin page, and you should no longer be able to access anything by entering in a new URL.


mshadbolt (approved):
confirmed working for me!

PR #672, opened 2024-07-10
Keycloak prod

SonQBChau:
## What's New
- Updated Keycloak settings for production.
- This PR assumes Keycloak is behind a reverse proxy, similar to the UHN setup, using HTTP traffic.
- Cleaned up and improved debug mode.

### How to Test
- Run the integration tests on docker
- For production, change the following

.env
```
CANDIG_PRODUCTION_MODE=1
KEYCLOAK_PUBLIC_PROTO=https
KEYCLOAK_PUBLIC_URL=${KEYCLOAK_PUBLIC_PROTO}://${CANDIG_AUTH_DOMAIN}
KEYCLOAK_PRIVATE_URL=${KEYCLOAK_PRIVATE_PROTO}://keycloak:${KEYCLOAK_PORT}
```

docker-compose
```
KC_PROXY_HEADERS: xforwarded # FOR PROD
```


etc/env/example.env
```diff
@@ -165,12 +165,12 @@ KEYCLOAK_REALM=candig
 KEYCLOAK_CLIENT_ID=local_candig
 KEYCLOAK_LOGIN_REDIRECT_PATH=/auth/login
 KEYCLOAK_PORT=8080
+KEYCLOAK_CONTAINER_PORT=8080
```
daisieh: Why do we need a separate entry for the container port?

lib/keycloak/client_setup.sh
```diff
@@ -13,25 +11,19 @@ CREATE_OUTPUT=$(KCADM create clients -r "$KEYCLOAK_REALM" \
...
 
-# EXPORT: Get the client secret and save it to secrets
-# echo $CLIENT_ID
-# echo $(KCADM get clients/"$CLIENT_ID"/client-secret -r "$KEYCLOAK_REALM")
-CLIENT_SECRET=$(docker exec "$KEYCLOAK_CONTAINER_ID" /opt/keycloak/bin/kcadm.sh get clients/"$CLIENT_ID"/client-secret -r "$KEYCLOAK_REALM" | sed -n '/{/,/}/p' | jq -r '.value')
-# echo $CLIENT_SECRET
-echo "$CLIENT_SECRET" > tmp/keycloak/client-secret
+# EXPORT client secret
+CLIENT_SECRET=$(KCADM -full get clients/"$CLIENT_ID"/client-secret -r "$KEYCLOAK_REALM" | sed -n '/{/,/}/p' | jq -r '.value')
+echo -n "$CLIENT_SECRET" > tmp/keycloak/client-secret
```
daisieh, line 29: I never knew there was a -n option for echo! This is life-changing.

lib/keycloak/credential_loader.sh
```diff
@@ -3,4 +3,8 @@
 # Load credentials from secrets
 export KEYCLOAK_ADMIN_PASSWORD=$(< /run/secrets/keycloak-admin-password)
 
-exec /opt/keycloak/bin/kc.sh "$@"
\ No newline at end of file
+if [ "$CANDIG_DEBUG_MODE" = 1 ]; then
```
daisieh: We should create a `CANDIG_PRODUCTION_MODE` entry in example.env and then use that here instead, since it's possible (maybe?) to want to debug a production server.

daisieh (commented):
Tiny comments, overall looks good.

etc/env/example.env
SonQBChau: It was added on prod, thought we need it

SonQBChau:
- remove `KEYCLOAK_CONTAINER_PORT `
- add `CANDIG_PRODUCTION_MODE `

daisieh (approved):
LGTM

PR #673, opened 2024-07-11
Remove manual ingest from nightly build

mshadbolt:
* Removes ingest of test files from ingest because the build keeps the integration test data now
* Ensures keep_test_data is true in the .env file


daisieh (approved):
LGTM! I'm assuming this is the branch that is working correctly on dev right now?

mshadbolt:
Yep I used it for the last successful build

PR #685, opened 2024-08-16
Logging tweaks

daisieh:
Small changes:

* added logging module to etc/venv/requirements.txt
* added DEBUG_MODE to more docker-composes
* updated clean-containers target to delete the logging container last, as deleting it early messes up the rest of the containers

To test: should be able to do a full build-all/clean-all/build-all and have that work out.


OrdiNeu (approved):
LGTM :+1:

PR #693, opened 2024-08-16
Update print_logs.sh

daisieh:
container_logs.txt now contains the output of the fluentd log.


mshadbolt (approved):
checked file exists and contains the formatted logs, LGTM!

PR #705, opened 2024-08-21
Update to model 3

SonQBChau:
## What's new
- Include the develop branch of each submodules for model 3
- Update test-integration

### Hash commit for each submodules
Entering 'lib/candig-data-portal/candig-data-portal'
4fd58864546f3f8e886ce5e45e6320a743d96701
Entering 'lib/candig-ingest/candigv2-ingest'
035f8fc2f1ad68e2b8f89a55470b0a01c621d101
Entering 'lib/federation/federation'
3ce89391443fdec209b740b4d20876ca4e705ed8
Entering 'lib/htsget/htsget_app'
b6e01f920e2ec62833f601fcca97b8d8786c5c56
Entering 'lib/katsu/katsu_service'
10409345385b1bc8b1d4b38d795f699a59dc4314
Entering 'lib/opa/opa'
51669ae1f39e1a9e23f086bf73b6be8b3bed5471
Entering 'lib/query/query'
9f13d806878121d11c27e3ff3ab18b3eed88547d


etc/env/example.env
```diff
@@ -14,7 +14,7 @@ CANDIG_AUTH_DOMAIN=candig.docker.internal
 CANDIG_SITE_LOCATION=UHN
 CANDIG_DEBUG_MODE=1
 CANDIG_PRODUCTION_MODE=0
-CANDIG_VERSION=v4.1.0
```
mshadbolt: I thought we normally wait to update the versions when we do stable releases, this version doesn't actually exist yet, same for the katsu and ingest versions

mshadbolt (commented):
Deployed locally and on UHN dev server and all tests are passing.

Only comment was about incrementing to versions that don't exist yet in the .env file, but also not sure where that came from

etc/env/example.env
SonQBChau: ok, will roll back

mshadbolt:
Found a few bugs from the front-end but don't think it should prevent merging this to develop

mshadbolt (approved):
LGTM

OrdiNeu (approved):
LGTM :heavy_check_mark:

daisieh (approved):
Looks good to me. Builds and passes tests on my machine.

CourtneyGosselin (approved):
LGTM builds and passes on my machine (WSL windows) Found an issue on the portal but its an improvement making a ticket

PR #708, opened 2024-08-23
rename metadata to postgres

SonQBChau:
## What's new
- Rename `metadata` to `postgres`

### How to test
- Checkout `rename_db` branch for katsu
- Checkout `rename_db` branch for htsget
- Rebuild and run test integration


mshadbolt (approved):
Needed to manually delete the old metadata-db container for this to work since clean-all will not delete it anymore, we should add that to the release notes so that users know they need to do this when updating to this release.

daisieh (approved):
Works for me!

PR #717, opened 2024-08-28
change db name to clinical

SonQBChau:
- Katsu db name is still metadata, this should update it


mshadbolt (approved):
confirmed integration tests passed and checked the postgres container, confirmed clinical exists and metadata does not

PR #718, opened 2024-08-30
DIG-1720: Clinical data ingest is asynchronous

daisieh:
Creates a separate ingest daemon process that will do the actual Katsu data ingest.

When a user calls the ingest API endpoint, the json input is validated and pre-processed into per-program chunks. Those programs are then authorized against the logged-in user. If the user is authorized to ingest the programs and the input json is valid, the pre-processed json is saved to a temp file and given a queue ID, which is returned to the user.

There is now a new endpoint, `/status/<queue-id>`, that allows the user to check to see if the ingest has completed, and if so, what the result was. This endpoint does not check for authorization (except the usual Tyk check), since it wouldn't be accessible to anyone without the queue ID and does not give any clinical data in the results.

The clinical ingest tests in integration tests now use these two endpoints.

### DIFF
--- a/etc/tests/test_integration.py
+++ b/etc/tests/test_integration.py
@@ -442,7 +442,7 @@ def test_ingest_not_admin_katsu():
 
     response = requests.post(f"{ENV['CANDIG_URL']}/ingest/clinical", headers=headers, json=test_data)
     # when the user has no admin access, they should not be allowed
-    assert response.status_code == 401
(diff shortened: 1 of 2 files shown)


SonQBChau (approved):
Works well with no issues. I think the sleep time could be reduced to 1 second, but it may not be significant time saving anyway.

PR #731, opened 2024-09-09
Other/performance testing

OrdiNeu:
Wrote a small script to allow endpoint testing, callable from etc/tests/test-endpoints-all.py

Requires a file like the following:
[endpoints.csv](https://github.com/user-attachments/files/16933621/endpoints.csv)

Call it with:
```
python settings.py
source env.sh
python etc/tests/test-endpoints-all.py --token $REFRESH_TOKEN  --secret $CANDIG_CLIENT_SECRET --in endpoints.csv --out ~/Documents/timing/
```

or the approriate URL/variables for the candig-dev server.

Sample timings are available here: https://docs.google.com/spreadsheets/d/1lJd7JdfczPC7nqDy4vqTWpqHJoKN2svG4uDFLhX56gU/edit?gid=693470897#gid=693470897


etc/tests/test-endpoints-all.py
```diff
@@ -0,0 +1,84 @@
...
+        extra = split[1:]
+        sanitized_url = re.sub(r'https?://', "", url)
+        sanitized_url = re.sub(r'/', "_", sanitized_url)
+        token = get_token(args.username, args.password, args.secret, args.loginurl)
+        print(url)
+
+        rc = subprocess.run(
+            ['python', os.path.join(os.path.dirname(os.path.realpath(__file__)), 'test-endpoint.py'),
+            '--url', url,
+            '--out', args.out + sanitized_url + ".csv",
```
daisieh: maybe do a proper dir join for this?

daisieh (changes requested):
Would it be easier to just use the `site_admin_token.py` script to get the token, instead of having the user pass in the site admin credentials, login url, secret?

OrdiNeu:
@daisieh 
> Would it be easier to just use the `site_admin_token.py` script to get the token, instead of having the user pass in the site admin credentials, login url, secret?

That would force the user to only be able to test their own setup, rather than letting e.g. a user test candig-dev from outside.

daisieh:
Oh right; that makes sense. It's just not very user-friendly to have to pass in so many arguments...would it be easier to have the user pass in a refresh token only?

daisieh (approved):
Looks good, works for me!

PR #732, opened 2024-09-09
If token is bad, remove cached token

daisieh:
If the refresh token is bad, it'll error out and you need to run the script again. This is annoying. Delete any bad cached refresh token before trying again.

### DIFF
--- a/site_admin_token.py
+++ b/site_admin_token.py
@@ -37,7 +37,10 @@ def get_site_admin_token(username=None, password=None, refresh_token=None):
             )
 
         if "error" in credentials:
-            os.remove("tmp/site-admin-refresh-token")
+            try:
+                os.remove("tmp/site-admin-refresh-token")
+            except:
+                pass
             return get_site_admin_token()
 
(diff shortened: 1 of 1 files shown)


site_admin_token.py
```diff
@@ -37,13 +37,15 @@ def get_site_admin_token(username=None, password=None, refresh_token=None):
...
         if "error" in credentials:
+            os.remove("tmp/site-admin-refresh-token")
             return get_site_admin_token()
 
         with open(f"tmp/site-admin-refresh-token", "w") as f:
             f.write(credentials["refresh_token"])
 
         return credentials["access_token"]
     except Exception as e:
+        os.remove("tmp/site-admin-refresh-token")
```
OrdiNeu: If this crashes (e.g. because the file does not exist), it'll crash out the call to `get_site_admin_token()`. Recommend just placing a try block around this so it continues even if it fails.

OrdiNeu (approved):
Untested but looks good :heavy_check_mark:

PR #737, opened 2024-09-12
Allow Katsu to use refresh tokens as well as access tokens

OrdiNeu:
# Description
This PR allows Katsu to accept refresh tokens when queried directly, in addition to the usual access tokens.
This builds off of work done in #709 

# To Test
Grab the refresh token from:
```
curl -s --request POST \                                                                                                                                                                                               
   --url 'http://candig.docker.internal:8080/auth/realms/candig/protocol/openid-connect/token' \                                                                                                                                      
   --header 'Content-Type: application/x-www-form-urlencoded' \                                                                                                                                                                       
   --data grant_type=password \                                                                                                                                                                                                       
   --data client_id=local_candig \                                                                                                                                                                                                    
   --data client_secret=$CANDIG_CLIENT_SECRET \                                                                                                                                                                                       
   --data username=$CANDIG_SITE_ADMIN_USER \                                                                                                                                                                                          
   --data password=$CANDIG_SITE_ADMIN_PASSWORD \                                                                                                                                                                                      
   --data scope=openid
```

Then plug it in the below:

```
curl http://candig.docker.internal:8008/v3/authorized/donors/ -H "Authorization: Bearer $TOKEN"
```

It should work with both the refresh&access tokens now.

### DIFF
--- a/lib/katsu/katsu_setup.sh
+++ b/lib/katsu/katsu_setup.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -Euo pipefail
+
+LOGFILE=$PWD/tmp/progress.txt
+
+# This script runs after the container is composed.
+
+echo ">> waiting for katsu to start"
+ingest=$(docker ps --format "{{.Names}}" | grep katsu_1)
+while [ $? -ne 0 ]
+do
+  echo "..."
+  sleep 1
+  ingest=$(docker ps --format "{{.Names}}" | grep katsu_1)
+done
+sleep 5
+
+
+bash $PWD/create_service_store.sh "katsu"
+
+docker restart $ingest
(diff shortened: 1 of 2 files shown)


lib/katsu/katsu_setup.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -Euo pipefail
+
+LOGFILE=$PWD/tmp/progress.txt
+
+# This script runs after the container is composed.
+
+echo ">> waiting for katsu to start"
+ingest=$(docker ps --format "{{.Names}}" | grep katsu_1)
```
daisieh: might want to change this var name to `katsu` here and below

daisieh (approved):
LGTM but tiny change requested.

SonQBChau (approved):
LGTM

OrdiNeu:
Test failure is an ingest failure which I assume is the stochastic one that shows up every now and again?

PR #748, opened 2024-09-19
DIG-1778: Improve timing of Query calls

OrdiNeu:
# Jira Link
[DIG-1778](https://candig.atlassian.net/browse/DIG-1778)

# Description
This creates a new `/v3/authorized/query/` endpoint in Katsu to drastically improve the performance of Query by combining all of its required information&filtering into one request. No changes to the output should be present, but the performance should see a ~2x speedup.

# To test
There are unit tests in Katsu under `lib/katsu/katsu_service/chord_metadata_service/mohpackets/tests/endpoints/test_donor.py` but mostly the integration tests should catch most of this. No change to Query should be present, besides the speedup.

# Other PRs
[Katsu](https://github.com/CanDIG/katsu/pull/266)
[Query](https://github.com/CanDIG/candigv2-query/pull/52)

[DIG-1778]: https://candig.atlassian.net/browse/DIG-1778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SonQBChau (approved):
I tested and it works fine. But, it took me a while to manually override all the synthetic data for ingestion, so I recommend holding off on merging until we have the updated synthetic data.

PR #750, opened 2024-09-20
Use internal URL for Query->OPA

OrdiNeu:
# Jira ticket
[DIG-1780](https://candig.atlassian.net/browse/DIG-1780)

# Description
Fixup Query to use the private OPA URL instead of the public one. This is the only instance of public urls being used where it could be switched to private ones I could find in all the docker-composes -- data-portal needs public URLs as it's run from the client. Ingest does use a public KEYCLOAK/CANDIG_URL for some reason but I assume that was necessary?

[DIG-1780]: https://candig.atlassian.net/browse/DIG-1780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


daisieh:
I just did a quick grep on the candig-ingest repo and KEYCLOAK_PUBLIC_URL doesn't ever seem to be used...it could and probably should be removed.

daisieh:
The CANDIG_URL is not called directly in the container setting, but is needed for making DRS addresses and for reference to the user in certain settings, so that one should be left alone.

daisieh:
Francis pointed out that KEYCLOAK_URL is required for authx, and is not actually going through Tyk anyway, so that should be left alone as well.

daisieh (approved):
LGTM!

PR #752, opened 2024-09-21
add recompose target

daisieh:
I got tired of doing make clean/docker-volumes/build/compose, so I added a convenience target that does all of them in one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -294,13 +294,12 @@ compose-%:
 
 
 #>>>
-# NOT the same as make clean-compose! This is a shortcut that combines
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -293,6 +293,12 @@ compose-%:
 	echo "    finished compose-$*" >> $(LOGFILE)
 
 
+cleancompose-%:
```
OrdiNeu: Might want a comment above this explaining what it does

OrdiNeu (approved):

OrdiNeu (approved):

PR #754, opened 2024-09-23
DIG-1648: UHN Production deployment login page doesn't have candig favicon

OrdiNeu:
# Description

This adds a barebones theme onto the Keycloak login page with a modified favicon -- CanDIG's favicon

# To test
1. Build this branch
2. Go to the login page
3. The favicon at the top left should be the CanDIG one instead of the Keycloak default
image


mshadbolt (approved):
works for me

PR #760, opened 2024-09-24
Save logs to a different logs directory

daisieh:
I just realized that saving the logs to a directory called `tmp/logging` means that the `make clean-logging` target will delete the logs every time we clean-all. It would be safer to save them to a differently-named directory; I've chosen `tmp/logs`.


mshadbolt (approved):
lgtm

PR #766, opened 2024-09-26
DIG-1769: backup and restore vault

daisieh:
To test: make sure you've added something extra to your existing vault setup, like adding user1 to the site curators:
```
curl -X "POST" "http://candig.docker.internal:5080/ingest/site-role/curator/email/user1@test.ca" \
     -H 'Authorization: Bearer <site admin token>' \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d $'{}'
```

Then run the target `make backup-vault`. The backup file will be in tmp/vault/backup.tar.gz.

To restore, run `make clean-vault` and then copy that backup file to `lib/vault/restore.tar.gz`. Then run `make restore-vault`. The backup file will be renamed to `lib/vault/restored.tar.gz` so that you can remove it or not at your leisure.

Then when you check your site curators, user1 should still be there:
```
curl "http://candig.docker.internal:5080/ingest/site-role/curator" \
     -H 'Authorization: Bearer <site admin token>' \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d $'{}'
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -561,3 +561,10 @@ rebuild-keep-data:
 	# Rebuild everything
 	$(foreach MODULE, $(CANDIG_MODULES), $(MAKE) build-$(MODULE); $(MAKE) compose-$(MODULE);)
 	./post_build.sh
+
+
+# wrapper for make_backup.sh to make sure we're running it from the right directory
+backup-vault:
+	@bash lib/vault/make_backup.sh
+	-$(MAKE) compose-vault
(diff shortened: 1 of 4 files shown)


mshadbolt:
Need to move the backup file before running make clean-vault because make clean-vault deletes the file

mshadbolt (approved):
Works for me! will document in the documentation PR

PR #767, opened 2024-09-26
Fix candig-ingest submodule commit

mshadbolt:
Accidentally committed with a commit hash from a branch that no longer exists, this changes it to the latest develop commit.


daisieh (approved):
LGTM

PR #770, opened 2024-09-27
DIG-1712: Ingest should prevent ingest if program auth doesn't exist

daisieh:
Added tests to make sure that program auths exist before clinical ingest.


mshadbolt (approved):
LGTM

PR #772, opened 2024-09-30
update ingest to develop commit

mshadbolt:
* Update ingest submod to the merged ingest commit instead of the one from deleted branch


daisieh (approved):
LGTM

PR #773, opened 2024-09-30
DIG-1770: stack should stop and start containers in the same order

daisieh:
On `make build-all`, we start a file at tmp/containers.txt to record all of the containers that get started up. As each `make compose-%` target runs, we add the containers that are created to the tmp/containers.txt file. (We make sure that any containers that are already listed don't get added again.)

Then, when we run `make stop-all`, we list out the containers in reverse order and stop in that order. On `make start-all`, we start the containers in their order of creation. We give a small pause to make sure that each container get started before the next one gets started.


mshadbolt (approved):
works for me!

PR #776, opened 2024-10-01
DIG-1397: Remove check for UHN VPN

mshadbolt:
This completely removes the check.

### DIFF
--- a/pre-build-check.sh
+++ b/pre-build-check.sh
@@ -1,12 +1,5 @@
 #!/usr/bin/env bash
 
-# Check 1: UHN VPN is up
-if command -v getent >/dev/null 2>&1; then
-    CANDIG_HOST=`getent hosts candig-dev`
-elif command -v dscacheutil >/dev/null 2>&1; then
-    CANDIG_HOST=`dscacheutil -q host -a name candig-dev`
-fi
-
 SILENT_MODE=0
 if [[ $* == *-s* ]]; then
     SILENT_MODE=1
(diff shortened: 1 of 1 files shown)


pre-build-check.sh
```diff
@@ -1,12 +1,5 @@
 #!/usr/bin/env bash
 
-# Check 1: UHN VPN is up
```
daisieh, line 3: Seems like it'd be easier to take the clause looking for $CANDIG_HOST out of the if statement below, rather than take away the calculation?

daisieh (changes requested):

pre-build-check.sh
mshadbolt: removed the if statement too

daisieh: leave in the $SILENT_MODE check...that is still useful.

daisieh (changes requested):

pre-build-check.sh
mshadbolt: I don't get what you mean, the silent mode is just to check whether to print the message about the uhn vpn or not, but we aren't doing that anymore? Do you mean you want to do a silent mode check for the domain check too?

daisieh: I don't know; you're probably right. I think that my impulse is usually to remove as little code as possible, but perhaps that isn't very helpful. I'll go ahead and approve it.

daisieh (approved):

PR #779, opened 2024-10-02
Increase index time

mshadbolt:
Increases time to wait for indexing to complete

Adds auto deletion for tmp directory that gets created for generate_test_data script in ingest


OrdiNeu (approved):

OrdiNeu:
There's a failure in the test suite that I didn't catch

mshadbolt:
investigating

PR #796, opened 2024-10-04
increase default aggregate threshold

mshadbolt:
Increase the default aggregate threshold in the example env to 10


daisieh (approved):
LGTM

PR #811, opened 2024-10-11
hotfix: attach the proxy-headers value to CANDIG_PRODUCTION_MODE

daisieh:
Instead of using the env var, we can use the CLI flag when starting up in production mode.


SonQBChau (approved):
lgtm

CourtneyGosselin (approved):
Works on BCGSC-DEV

PR #812, opened 2024-10-15
DIG-1740: set channels to conda-forge

daisieh:
Instead of using default channels, use conda-forge.


mshadbolt:
Would be good to get Courtney to review to to make sure it works on WSL

CourtneyGosselin (approved):
Works on WSL and all integration tests pass

PR #816, opened 2024-10-18
hotfix: hardcode "federation" as the listen path for the federation service

daisieh:
To ensure that `federation` is always the listen_path for the federation service, hardcode the value and remove it from the env file.


OrdiNeu (approved):
LGTM

PR #820, opened 2024-10-18
Hotfix: handle if default site admin not present

daisieh:
Test by following the instructions in the doc below. Then uncomment out the DEFAULT_SITE_ADMIN_USER and do it again.


CourtneyGosselin (approved):
LGTM works on BCGSC-DEV

PR #826, opened 2024-10-22
Integration test data specific to Site location

mshadbolt:
Uses the env variable 'CANDIG_SITE_LOCATION' for synthetic data ids so that programs and other identifiers are unique to a site.

Changed clinical ingest to a single file ingest now that we have async we don't need to do programs one by one


daisieh (approved):
LGTM, works as expected

PR #837, opened 2024-10-25
Stable rebase

daisieh:
Merging v5.0.0 stable changes back to develop


.github/workflows/slack-notify,yaml
```diff
@@ -0,0 +1,23 @@
+name: Slack Notify on example.env Update
```
OrdiNeu, line 1: Did this not already exist in develop?

OrdiNeu (approved):
Worked for me

.github/workflows/slack-notify,yaml
daisieh: I have no idea why it wasn't...

PR #839, opened 2024-10-28
Bugfix: Nightly build was not actually storing/sending the git response when erroring out

OrdiNeu:
`'Could not automatically pull git repo: '` messages we get were all missing the reason the git repo could not be pulled, which are usually just that the branch does not exist.


daisieh (approved):
LGTM

PR #842, opened 2024-10-31
Remove residual refs to TYK_CANDIG_API

daisieh:
Removing unused file and references to deleted TYK_CANDIG_API. This shouldn't affect anything.


OrdiNeu (approved):
Untested but I see the build ran OK and the code does look unused :heavy_check_mark:

PR #843, opened 2024-10-31
Fixup the URL that's displayed at the end of Nightly Build

OrdiNeu:
The URL reported at the end of nightly build should probably be the login target URL.


daisieh (approved):
LGTM

PR #848, opened 2024-11-08
DIG-1427: delete data as well as auth when deleting a program

daisieh:


OrdiNeu (approved):
LGTM :heavy_check_mark:
This does look like it makes deleting programs a lot simpler, which is nice.

PR #859, opened 2024-11-13
Update api_ingest.json.tpl

daisieh:
If we're providing the url in the openapi schema in https://github.com/CanDIG/candigv2-ingest/pull/133, we need to make sure that the listen path is not getting stripped by tyk.


daisieh:
seems like this does not fix the problem. Closing in favor of https://github.com/CanDIG/candigv2-ingest/pull/137

PR #862, opened 2024-11-14
DIG-1812: store genomic completeness per cohort

daisieh:


OrdiNeu (approved):
I realize there's no completeness queries in the integration tests, but there should be -- I've written a new ticket for it (not related to this PR)

PR #870, opened 2024-11-19
DIG-1854: update clean-authx and init-authx

daisieh:
We haven't refactored clean/init-authx since we redid docker secrets and moved to vault secret stores. I think these changes should be sufficient to make things work again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -418,6 +418,7 @@ docker-volumes:
 .PHONY: init-authx
 init-authx: mkdir
 	$(MAKE) docker-volumes
+	$(MAKE) authx-secrets
 	$(foreach MODULE, $(CANDIG_AUTH_MODULES), $(MAKE) build-$(MODULE); $(MAKE) compose-$(MODULE); python settings.py;)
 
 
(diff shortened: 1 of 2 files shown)


mshadbolt (commented):
To test, started up the stack, ran integration tests and noted they passed.

Ran `make clean-authx`
Noted all authx modules are stopped and deleted

Ran `make init-authx`
Noted all authx modules get rebuilt and recomposed
Ran integration tests, the following tests failed:

3 federation tests fail: 
```
FAILED etc/tests/test_integration.py::test_services_count - TypeError: string indices must be integers, not 'str'
FAILED etc/tests/test_integration.py::test_federation_call - assert 'list' in "<class 'dict'>"
FAILED etc/tests/test_integration.py::test_add_server - KeyError: 0
```

Tried doing `make recompose-federation`

Then all tests passed

Not sure if federation always needs to be recomposed along with the authx modules?

mshadbolt (approved):
LGTM

daisieh:
Yep, federation always needs to be recomposed if Keycloak is recomposed: the issuer will have changed, so the server will have to be re-registered.

PR #875, opened 2024-11-20
DIG-1522: change references to "cohort" to "program"

daisieh:
HTSGet and, inconsistently, Katsu, use the old term "cohort" instead of the MoH v3 term "program." We need to update this.


OrdiNeu (approved):

OrdiNeu:
With the caveat that data-portal might need some changes

OrdiNeu:
There's a few merge conflicts now

PR #877, opened 2024-11-21
DIG-1835: make a self-contained init-minio make target

daisieh:
I thought that the best way to achieve the goal of having an optional minio server would be to pull all of those bits out of the main make compose targets and consolidate them into one `init-minio` target. This way, if a user wants to set up a MinIO S3 store for testing or dev, they can easily do so. The access key and secret key are in tmp/secrets.

I removed the candigv2 label from the minio container/image/volumes because otherwise clean-all will grab these too. If you want to get rid of the minio container completely, `make clean-minio` works.

To test: run `make init-minio`. You should then have a minio container with its api visible at port 9000. If you run `make clean-all`, the minio container should remain and still be accessible. Running `make clean-minio` should get rid of it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,8 @@ pnpm-debug.log*
 
 # macOS-specific files
 .DS_Store
+
+# minio-related files
+lib/minio/access-key
+lib/minio/secret-key
+lib/minio/aws-credentials
--- a/Makefile
+++ b/Makefile
@@ -360,7 +360,7 @@ docker-push:
 
 #<<<
 .PHONY: docker-secrets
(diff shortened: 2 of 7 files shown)


mshadbolt:
Confirmed that minio targets function as specified. Confirmed that the minio container does not get deleted when running `make clean-all`, there is a problem though because `make clean-all` deletes all the secrets, including the minio keys.

mshadbolt (approved):
confirmed clean all no longer deletes minio credentials. LGTM

Did not test actual functionality of minio, lmk if you want me to try anything in that regard.

PR #878, opened 2024-11-21
Update ingest guide docs

mshadbolt:
* changed slugs for api docs in case we want to have descriptive service docs at some point
* updated katsu endpoint in ingest doc to `v3` so that it works properly
* small fixups


daisieh (approved):
LGTM

PR #880, opened 2024-11-22
DIG-1137, DIG-1618, DIG-1851: misc post-build script fixes

daisieh:
I consolidated all of the outputs into LOGFILE=progress.txt, since that file didn't seem to have anything useful in it anyway.

DIG-1137: only print out errors at the end for modules with errors or warnings
DIG-1618: remove orphan container warnings with env var COMPOSE_IGNORE_ORPHANS=true
DIG-1851: list the containers that are not running at the end. This is actually an improvement because I'm using the actual names of the expected vs running containers and not the counts, esp since additional containers like minio might be present.

On a running local build, you can test this by stopping various containers and then re-running `bash post_build.sh`. You should see the missing containers listed at the end.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,6 @@ SHELL = bash
 CONDA = $(CONDA_INSTALL)/bin/conda
 CONDA_ENV_SETTINGS = $(CONDA_INSTALL)/etc/profile.d/conda.sh
 
-LOGFILE = tmp/progress.txt
 
 .PHONY: all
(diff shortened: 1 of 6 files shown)


Makefile
OrdiNeu, line 1: On line 20 of this file the `LOGFILE` declaration over here might overwrite the variable defined in the .env

OrdiNeu (approved):

PR #894, opened 2024-11-29
hotfix: save temporary files

daisieh:
Save out temporary files in ingest and htsget in case the github test workflow fails in some way, so that we can inspect the files.

You can see the files in the post-build artifacts: https://github.com/CanDIG/CanDIGv2/actions/runs/12090019799/artifacts/2255285496


OrdiNeu (approved):
Unsure how to test but it looks good

Issue #901, opened 2024-12-05
Error response from daemon: Could not find the file /home/candig in container candigv2_opa_1


vedant-cd:
## Operating system
- Ubuntu 

## What version are you deploying
Linux (ubuntu 24.04)
## Step where failure occurred (make-?)
`make build-all`
```
create a kv store for opa
Error enabling: Error making API request.

URL: POST http://127.0.0.1:8200/v1/sys/mounts/opa
Code: 400. Errors:

* path is already in use at opa/

Error response from daemon: Could not find the file /home/candig in container candigv2_opa_1
Error response from daemon: Could not find the file /home/candig in container candigv2_opa_1
^Cmake[1]: *** [Makefile:292: compose-opa] Interrupt
make: *** [Makefile:100: build-all] Interrupt
```
## Are there errors in `tmp/error.txt`?
NO
## Are there errors in the central log for the affected container(s)? `tmp/logs/*.log`
NO

## Any other information

![image](https://github.com/user-attachments/assets/8ee032c1-2195-469a-a253-1d32ae9b37b3)

daisieh:
There are some error messages that occur because general scripts that we use to do setup steps across all containers don't completely apply to every container, but it's basically impossible to get Docker to not tell you when it doesn't work. This is an innocuous error message that is expected to occur.

PR #903, opened 2024-12-05
Only post output if error is unexpected

daisieh:
Inspired by #901, I decided to write a wrapper bash script that runs a command and compares it to the expected output. If the error that is outputted by the command is one that is expected, then it will just pass silently; it will only output the error if it was something unexpected.

This cleans up a few of the last expected-errors that we get in `make build-all`: if you build a clean stack, you shouldn't get any errors at all now.


CourtneyGosselin (approved):
LGTM works on my local

PR #911, opened 2024-12-18
Fixup test_add_remove_program_authorization to properly test different users

OrdiNeu:
I think there's a bug in this particular test -- the same user was being given auth & tested over and over.


daisieh (approved):
LGTM, passes tests

PR #912, opened 2024-12-18
Add debugging notes for the Quarkus error

OrdiNeu:
Added in a section to place specific errors that users might encounter, as well as debugging instructions for a specific misconfiguration that was run into at C3G.


mshadbolt (approved):
LGTM

PR #919, opened 2024-12-31
Update to opa v1.0.0

daisieh:
Updates the docker-compose file to match https://github.com/CanDIG/candig-opa/pull/76


OrdiNeu (approved):
Seems to work good :+1:

PR #931, opened 2025-01-07
Hotfix: bump authx to v2.6.2

daisieh:
Picks up change in https://github.com/CanDIG/candigv2-authx/pull/39


daisieh:
Marion tested these changes in https://github.com/CanDIG/candigv2-authx/pull/39.

PR #932, opened 2025-01-08
Updating authz tests to remove deprecated endpoint calls

daisieh:
In preparation for ingest changes to address https://candig.atlassian.net/browse/DIG-1874, I'm removing tests that call the deprecated endpoints. While looking at tests, though, I also streamlined some stuff: changed references to "datasets" to "programs" and switched the direct calls to OPA to use the authx library calls instead.


mshadbolt (approved):
Looks good to me and integration tests pass

PR #948, opened 2025-01-31
RNAGet (renamed from Transcriptomics) DIG-1878, DIG-1872

SonQBChau:
## What's new
- add rnaget into the stack
- integrate tds auth with OPA
### How to test
- update .env
- run submodule init
- rebuild the stack
- run test-integration

### DIFF
--- a/etc/tests/test_integration.py
+++ b/etc/tests/test_integration.py
@@ -76,6 +76,7 @@ def test_tyk():
     endpoints = [
         f"{ENV['CANDIG_ENV']['TYK_HTSGET_API_LISTEN_PATH']}/ga4gh/drs/v1/service-info",
         f"{ENV['CANDIG_ENV']['TYK_KATSU_API_LISTEN_PATH']}/v3/service-info",
+        f"{ENV['CANDIG_ENV']['TYK_RNAGET_API_LISTEN_PATH']}/service-info",
         f"federation/v1/service-info",
         f"{ENV['CANDIG_ENV']['TYK_OPA_API_LISTEN_PATH']}/v1/data/service/service-info",
         f"{ENV['CANDIG_ENV']['TYK_QUERY_API_LISTEN_PATH']}/service-info",
--- a/lib/rnaget/Dockerfile
+++ b/lib/rnaget/Dockerfile
@@ -1,0 +1,27 @@
+FROM ghcr.io/bento-platform/bento_base_image:python-debian-2024.07.09
+
+RUN apt-get update && apt-get install -y postgresql-client
+
+RUN pip install --no-cache-dir "uvicorn[standard]==0.30.1"
+
+
+WORKDIR /tds
+
+COPY pyproject.toml .
+COPY poetry.lock .
+
+RUN poetry config virtualenvs.create false && \
+    poetry --no-cache install --without dev --no-root
+
+COPY transcriptomics_data_service transcriptomics_data_service
+COPY lib lib
+COPY entrypoint.bash .
+COPY run.bash .
+COPY LICENSE .
+COPY README.md .
+COPY create_db.sh .
+
+RUN poetry install --without dev
+
+ENTRYPOINT [ "bash", "./entrypoint.bash" ]
+CMD [ "bash", "./run.bash" ]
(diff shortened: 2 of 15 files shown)


lib/rnaget/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+    environment:
+      - CORS_ORIGINS="*"
+      - BENTO_AUTHZ_SERVICE_URL=""
+      - OPA_URL=${OPA_PRIVATE_URL}
+      - VAULT_URL=${VAULT_PRIVATE_URL}
+      - KEYCLOAK_PUBLIC_URL=${KEYCLOAK_PUBLIC_URL}
+      - DB_HOST=postgres-db
+      - DB_PORT=5432
+      - DB_USER=${DEFAULT_ADMIN_USER}
+      - DB_NAME=tds_db
```
daisieh:
```suggestion
      - DB_NAME=rnaget_db
```

daisieh (approved):
I added one suggestion, that the db be named rnaget_db. Otherwise, it looks good.

etc/tests/test_integration.py
```diff
@@ -76,6 +76,7 @@ def test_tyk():
     endpoints = [
```
OrdiNeu, line 76: Recommend more tests for RNAget -- though will that be possible without the HTSGet middleware?

lib/rnaget/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN poetry config virtualenvs.create false && \
+    poetry --no-cache install --without dev --no-root
+
+COPY transcriptomics_data_service transcriptomics_data_service
+COPY lib lib
+COPY entrypoint.bash .
+COPY run.bash .
+COPY LICENSE .
+COPY README.md .
+COPY create_db.sh .
```
OrdiNeu, line 22: Can this be simplified? `COPY . /tds`?

lib/rnaget/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+      options:
+        tag: "{{.Name}}"
+    ports:
+      - "${RNAGET_PORT}:5000"
+    environment:
+      - CORS_ORIGINS="*"
+      - BENTO_AUTHZ_SERVICE_URL=""
+      - OPA_URL=${OPA_PRIVATE_URL}
+      - VAULT_URL=${VAULT_PRIVATE_URL}
+      - KEYCLOAK_PUBLIC_URL=${KEYCLOAK_PUBLIC_URL}
```
OrdiNeu: Are these two needed? I see OPA_URL is needed because of the `authx.auth.verify_service_token` call, but the other two don't look like they're used anywhere. Is this future-proofing in case we need more auth?

OrdiNeu (approved):
Looks good to me, unsure how to properly test it though. 

When looking at the endpoints listed in the `README.md`, most of them return 404s. In addition, the docs/ endpoint seems to not work.

image


A few other comments below:

etc/tests/test_integration.py
SonQBChau: Yes, more RNAget tests will need to go through ingest and htsget, but they are not ready yet.

lib/rnaget/Dockerfile
SonQBChau: There are extra files like .git, tests, and other Docker files that shouldn't be included. I copied this from their Docker setup, so I'm not sure what their original purpose not including them.

SonQBChau:
I investigated the docs bug a bit. The issue is that Tyk redirects the default path to the portal, so the resource isn't found at http://candig.docker.internal:5080/portal/openapi.json. If you want to access Swagger, you can use the internal link: http://candig.docker.internal:5005/docs. If you want it for production, you'll need to either copy the resource to the portal and serve it there or find a way to bypass Tyk's redirect.

lib/rnaget/docker-compose.yml
SonQBChau: Yeah, good catch. I originally had it there because I needed to use other endpoints as well. I can remove it.

mshadbolt (approved):
I am approving because I like the new name. I didn't do any extra testing but I think getting approval from Daisie and Francis is enough.

PR #952, opened 2025-02-03
Fixup the genomic tests so that the site is not hardcoded

OrdiNeu:
The site was hard-coded for some reason, throwing errors for the genomic tests on non-standard installs.


mshadbolt (approved):
lgtm thanks for the fix

PR #954, opened 2025-02-05
Use latest katsu develop to work with raw.githususercontent.com refs.

DavidBrownlee:
Katsu pulls the schema.yml from raw.githubusercontent.com with this:
"openapi_url": "https://raw.githubusercontent.com/CanDIG/katsu/develop/chord_metadata_service/mohpackets/docs/schemas/schema.yml"
When changes are made to the katsu develop schema.yaml, the latest katsu develop commit is needed so that the synthetic data is compatible with the raw.githubusercontent.com schema.


SonQBChau:
Hi David,

Could you make a PR on our katsu https://github.com/CanDIG/katsu so it would go through our review and PR process?

SonQBChau:
Oh, are you trying to merge the latest Katsu into CanDIG **stable**? Sorry, I didn't look closely. We have a couple of PRs to merge first, and it needs to go through katsu stable, so I'd be doing this later. Like I said, how soon do you need it?

SonQBChau:
We're planning another stable release this month. If you can wait, that would be better to avoid repeated testing and extra work.

DavidBrownlee:
Hi Son,
Thanks for taking a look at this.
I guess it can wait for the next stable build if it is only a few weeks.
This issue is that any and all deployed stable versions (BCGSC, UHN) are now broken.  Their make test-integrations will fail.
This is a patch to fix it.

mshadbolt:
Hi David, merging develop into stable is only something we would do for a stable release. I am working on a fix to how the integration tests pull the synthetic data that should fix the integration tests.

DavidBrownlee:
Hi Marion,
The problem appears to lie in using "openapi_url": "https://raw.githubusercontent.com/CanDIG/katsu/develop/chord_metadata_service/mohpackets/docs/schemas/schema.yml" as the validation schema since it breaks the atomaticity of the commit.  I believe we discussed this is the past.  In this case, the synthetic data follows one schema (stable v5.0.1) but the schema was updated to something different (from the develop branch of raw.githubusercontent.com).
Ideally, for a truly frozen stable commit (or any atomic commit), the synthetic data and the schema would come within the same repo and nothing would be ever be pulled in from outside.
Thank you for looking into this.

mshadbolt:
This PR provides a fix which you can try out if you like, waiting on review before merging but it passes integration tests on stable for our gh build https://github.com/CanDIG/CanDIGv2/pull/958

PR #955, opened 2025-02-05
David/preserve test integration logs

DavidBrownlee:
Save a copy of the make test-integration output to ./tmp/test.


mshadbolt:
I think this is a good idea but we don't generally make changes to stable, can you make a new PR against develop and we can incorporate it into our next stable release. Thanks.

PR #956, opened 2025-02-05
Preserve make test-integration logs to tmp/test directory

DavidBrownlee:
Save a copy of the make test-integration output to ./tmp/test.

To test:
- ensure requirements are updated by reinstalling or specifically running `pip install pytest-reportlog`
- run `make test-integration`
- check a jsonl log file is created in `tmp/test/` with the date and time the test was run and the results of the tests as json objects in the file


mshadbolt:
I like the idea of this but it interferes with pytest. It makes it so the text colour highlighting is no longer present which is relatively minor. But the thing I find a dealbreaker is that normally, if you exit the tests early beacuse something is failing, before this change, it would show you the detailed breakdown for the tests that failed, with this change, it just exits and you don't get that information. Is there a way to do this that doesn't mess with the native behaviour of pytest?

mshadbolt:
e.g. before change:
Screenshot 2025-02-07 at 3 54 24 PM
Screenshot 2025-02-07 at 3 54 33 PM

after change:
Screenshot 2025-02-07 at 3 54 59 PM

DavidBrownlee:
Hi Marion,
All valid points - thank you for pointing them out.
Changes have been made to now preserve colours on the command line and the full stack trace even upon keyboard interrupt.
The only slight glitch is that the test-integration stored log output is saved with colours (rather than as raw text file), so depending you view it, it can appear a bit mashed up (as in the picture).  But it is better saved somewhere, somehow than not.

Screenshot from 2025-02-13 17-06-05

mshadbolt:
This change didn't work for me, got an error to do with the `-c` option that was used with `script`. I found a plugin that automatically captures test output as a json lines file https://pypi.org/project/pytest-reportlog/ and implemented that and it seems to be working

mshadbolt:
Confirmed working by @SonQBChau as well. Readability isn't the best but we can think about that and implement something in the future

mshadbolt (approved):
working after updates

PR #958, opened 2025-02-05
Hotfix: Use stable branch of synth data for integration tests

mshadbolt:
Uses the stable branch of mohccn data repo for integration tests by pulling a newer version of the generate_test_data.py from ingest

### DIFF
--- a/lib/candig-ingest/candig-ingest_setup.sh
+++ b/lib/candig-ingest/candig-ingest_setup.sh
@@ -21,4 +21,4 @@ bash $PWD/create_service_store.sh "candig-ingest"
 
 docker restart $ingest
 
-python $PWD/lib/candig-ingest/candigv2-ingest/generate_test_data.py --tmp tmp/data/synthdata --delete
+python $PWD/lib/candig-ingest/candigv2-ingest/generate_test_data.py --tmp tmp/data/synthdata --branch "stable" --delete


DavidBrownlee:
I can't explain why this patch works, but it doesn't appear to solve the inherent problem that commits are still not atomic and will break as drift occurs over time.

Within the candigv2-ingest submodule generate_test_data.py is the line:
synth_repo = Repo.clone_from("https://github.com/CanDIG/mohccn-synthetic-data.git", args.tmp, branch=args.branch)

Wouldn't the mohccn-synthetic-data repo be better included in the requirements.txt and pinned to a specific tag or commit (not a drifting branch)?

mshadbolt:
If people deploy stable, things are not supposed to change until the next stable (assuming we haven't made mistakes and need to hotfix, which we have this time) and the mohccn data stays in sync with stable I don't see the issue? But probably I need to think about it more, this was a quick fix with hopefully minimal impacts to anyone who has already deployed.

DavidBrownlee:
If I understand this correctly, this is a patch for stable, but not one that can ever be put into develop.  Develop will require the develop branch of the mohccn-synthetic-data repo (or something suitably more up to date).
So while a patch for this stable, this won't be in develop and will probably be absent in the next stable release (unless someone remembers to update this particular piece of code for all future stable releases) and things will break again.
I am not familiar with the synthetic data generation and ingestion testing, but does this accurately describe the situation?

mshadbolt:
it is not a perfect solution, but adding it as a requirement wouldn't work because it isn't a package. I'll have a think about how we can incorporate it in a more robust way but yes, you are right that the stable branch will always need to use 'stable' for the branch argument for the 'generate_test_data.py' when it is run in the ingest-setup file, otherwise it will fail (if the katsu model changes).

The patch doesn't need to be in develop because it pulls the develop version of the synthetic data by default anyway.

daisieh (approved):
LGTM

PR #959, opened 2025-02-06
DIG-1934: test-tyk only runs on listed CANDIG_MODULES

daisieh:
I had to add a bit of exception handling as well: in the edge case that you've built without rnaget but then put it into the env file, rnaget's service-info will redirect to the login page. That isn't a valid service-info response, so I added a check to make sure that the response you're getting is in fact a JSON, which is what service-info should return.


SonQBChau:
Tested, everything looks good! ✅

SonQBChau (approved):

PR #963, opened 2025-02-11
hotfix test for uhn-dev: try without https if ssl error

daisieh:
UHN's dev machine hosts the minio server at candig-demo.uhndata.io:9000. When tests are run on that machine, we need to use http instead of https. Since I don't know of any other constantly-accessible S3 servers that use publicly-available credentials, I hard-coded a retest to try http in case the failure is SSLError.


SonQBChau (approved):
LGTM

PR #971, opened 2025-02-14
Run Git Workflow on pushes to both branches develop and stable.

DavidBrownlee:
Runs the build and integration tests on pushes to both develop and stable branches.


mshadbolt (approved):
LGTM

PR #985, opened 2025-03-12
DIG-1951: Federation needs to check for candig-authorization

daisieh:
Federation needs to check for candig authorization itself and only send the fanout call to other federation nodes if candig-authzed locally.
If a federation node gets a call from another server's fanout call, we know that it's from a candig-authzed user on that node, so federation can add its service token before sending the call to its locally-registered services.
When the query service gets discovery calls from federation, it should check for the federation service token, and if it is there, it should allow the call.


daisieh:
Closing because outdated

PR #987, opened 2025-03-17
Allow Keycloak to use proxy headers in development mode

OrdiNeu:
Allow `KEYCLOAK_PROXY_HEADERS` to work even if we're in development mode.


mshadbolt:
Is this something we want to put into stable as well?

SonQBChau (approved):
LGTM

PR #989, opened 2025-03-17
Allow Keycloak to use proxy headers in development mode

OrdiNeu:
Per Son's request, this PR is going into `stable-candidate` instead of develop


SonQBChau (approved):
LGTM

PR #990, opened 2025-03-18
Upgrade Tyk to latest version 5.5.2

daisieh:
We can consider whether or not we need to reimplement healthcheck later.


OrdiNeu (approved):
Worked even after rebasing off of `develop` :heavy_check_mark:

PR #1002, opened 2025-03-26
Changing the realm caused integration tests to fail

OrdiNeu:
If you change KEYCLOAK_REALM in the `.env`, the integration tests were all failing. This should fix that.


daisieh (approved):
Yep, makes sense. LGTM.

PR #1003, opened 2025-03-27
Add nightly federation support

OrdiNeu:
By setting up a cron job for `auto_federate.py` that occurs after both servers are sure to have finished setting up, we can scan through all messages from today in the #automation-email channel, and automatically federate to them.

Will convert out of draft once I prove that it works for one day


daisieh (approved):
LGTM

PR #1016, opened 2025-04-01
hotfix: need to put comments on later lines

daisieh:
Make seems to keep the spaces as part of the string value; we need to put the comments on the next line.


CourtneyGosselin (approved):
LGTM

PR #1023, opened 2025-04-09
Fixup an issue where changing the Keycloak realm caused Keycloak_setup to crash out

OrdiNeu:
This didn't affect Keycloak itself working, but did affect the themed page not showing up.


mshadbolt (approved):
lgtm

PR #1025, opened 2025-04-14
Reorder the response so that a failing federate string doesn't crash out the NB. Also report on what the federate string should be for debugging purposes

OrdiNeu:
NB's been randomly failing last week. Reordering the messages should make it more fault-tolerant, and echoing the output should help with future debugging


daisieh (approved):
LGTM

PR #1026, opened 2025-04-14
DIG-1993: add a pause after restart during ingest setup

daisieh:
After the service store is set up, the setup script restarts the ingest container, but does not wait to make sure that the container is up and responding before trying to go on and use the API.

### DIFF
--- a/exec_with_expected.sh
+++ b/exec_with_expected.sh
@@ -9,6 +9,6 @@ grep -q "$expected" .tmp_store
 
 if [[ $? -ne 0 ]]; then
     cat .tmp_store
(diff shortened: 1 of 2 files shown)


OrdiNeu:
Will approve after the check finishes, but idly wondering here: would it be more accurate to check the return of /service-info? Might be a more accurate health check for when it has completed setup.

daisieh:
Actually yeah, that is probably a better idea.

OrdiNeu (approved):
The new code looks neat and clean :heavy_check_mark: Just waiting on the check to pass

PR #1033, opened 2025-04-17
Remove colour from the NB's output

OrdiNeu:
Quick fix to the error that NB sometimes encounters -- when it runs integration tests, sometimes the color-output that pytest outputs will mess up its output.


daisieh (approved):
LGTM, but didn't test

PR #1034, opened 2025-04-17
Add federation tests

OrdiNeu:
# Description
This introduces two new kinds of tests, run under make: `make test-local-federation` and `make test-querying-federation`. One site must run the "local" set of tests, which sets up datasets. The other must run the "querying" tests, which checks permissions on those datasets.

## To Test
There are two servers are UHN that currently use it, `candig-federation-1` and `candig-dev`. They should run nightly build first, then nightly federation, and then these tests. `candig-federation-1` runs its tests first and is setup to be the "local" set of tests that sets up a dataset to be queried by the other server. `candig-dev` is the "querying" server and runs those tests.


daisieh (approved):
I'm assuming that the fact that this works on our dev systems means that the code works. Otherwise, LGTM

PR #1043, opened 2025-05-01
DIG-1913: Allow ingest of transcriptomic matrices to htsget

daisieh:
picks up the latest commit for synthetic data as well.


SonQBChau (approved):
LGTM and passed integration test 👍

PR #1046, opened 2025-05-05
Bugfix/unremovable conda defaults

OrdiNeu:
# Description
As of last week, the latest version of conda appears to no longer allow you to remove the defaults channel without erroring out, which causes the rest of the `bin-conda` command to fail. Fresh installs will therefore be unable to function. However, I've gotten the build to work without the command, so I've just commented it out here.

# To Test
Reinstall your conda via bin-conda and init-conda under this branch. Instead of erroring out at removing the default channel, it should continue as normal.


OrdiNeu:
Underlying cause fixed in https://github.com/CanDIG/CanDIGv2/commit/b105431c2904de8a11224ac350c250a5e59f8ec9

PR #1060, opened 2025-05-14
Add playwright testing

SonQBChau:
## What's new
- Added Playwright tests to the GitHub workflow

### How to test
- Push any change (e.g., create a test.txt file) to trigger the GitHub Action (will change it back once done)
- Only the summary tests. Other tests are being fixed
- Playwright results can be found in Github Action Artifacts


.github/workflows/candig-testing.yml
```diff
@@ -90,3 +83,38 @@ jobs:
                 tmp/vault_audit.log
                 tmp/ingest
                 tmp/index
+
+      - name: Clone CanDIG Playwright Runbook
```
mshadbolt, line 95: I think it would be good to add all the Playwright tests as a separate job rather than all under Build-CanDIG

mshadbolt (changes requested):

mshadbolt (approved):
lgtm

.github/workflows/candig-testing.yml
SonQBChau: Playwright test can't be split into a separate task since it relies on the CanDIG instance from the workflow

mshadbolt (approved):

PR #1063, opened 2025-05-15
[DHDPS-333] Allow the Keycloak /auth prefix to be configurable from the .env file

OrdiNeu:
# Description
Allow the Keycloak URL's `/auth` prefix to be configured via the `.env` file. This was an optional prefix that we had enabled in order to maintain backwards compatibility, but must be removed in certain production instances (e.g. DHDP's prod)

# Other PRs
- [AuthX](https://github.com/CanDIG/candigv2-authx/pull/45)


etc/env/example-production.env
```diff
@@ -177,6 +177,7 @@ KEYCLOAK_PUBLIC_URL=${KEYCLOAK_PUBLIC_PROTO}://${CANDIG_AUTH_DOMAIN}
 KEYCLOAK_PRIVATE_URL=${KEYCLOAK_PRIVATE_PROTO}://keycloak:${KEYCLOAK_PORT}
 KEYCLOAK_REALM_URL=${KEYCLOAK_PUBLIC_URL}/auth/realms/${KEYCLOAK_REALM}
```
daisieh: do these need to be updated to the prefix too?

daisieh (approved):
LGTM, works as expected

PR #1064, opened 2025-05-15
DIG-1914: actually ingest rnaget data into Takuan

daisieh:
This is still using a feature branch of Takuan https://github.com/bento-platform/takuan/pull/29, but it allows us to ingest  transcriptomic data into Takuan.

Update: feature branch was merged as https://github.com/bento-platform/takuan/releases/tag/v1.0.0


SonQBChau (approved):
LGTM. I've added my only suggestion in the ticket https://candig.atlassian.net/browse/DIG-2039

but it can be another PR

PR #1068, opened 2025-05-23
Fixup typo in nightly federation that was causing it to always pass

OrdiNeu:
I think I figured out why it always reports success even when it really clearly shouldn't be


mshadbolt (approved):
lgtm

PR #1052, opened 2025-05-09
DHDPS-310: Store both the access and the refresh token in the cookie, and use either as needed

OrdiNeu:
# Other PRs
- https://github.com/CanDIG/candig-data-portal/pull/217

# Description
This changes how the session_id cookie is saved. Instead of just storing a refresh token, it stores `<refresh token>|<access token>`. The access token is refreshed by the user if needed, but is reused as much as possible, reducing the number of calls to Keycloak.

# Caveats
There's a couple caveats to the code I've written here:

1. I don't check the secret, because I couldn't figure out a way to store the public key of the token in a Tyk plugin.
2. This really only alters how cookie-based token usage works -- command line stuff should be mostly unaffected. The big improvement is that all of the requests from the frontend will no longer have to ping Keycloak for every request, as long as the tokens themselves are actually valid

PR #1069, opened 2025-05-26
CanDIG/candig-data-portal merging: DHDPS-310: Store both the access and the refresh token in the cookie, and use either as needed

candigbot:
PR triggered by update to develop branch on CanDIG/candig-data-portal. Commit hash: `530ec5ec0ae948d5594f594cd45900c96c506b48`. PR link: [#217](https://github.com/CanDIG/candig-data-portal/pull/217)


OrdiNeu:
Already included in #1052

PR #1070, opened 2025-05-27
DIG-2028: add crontab to rerun vault_setup every month

daisieh:
Quick and dirty fix: install a crontab to rerun vault setup every month.

### DIFF
--- a/lib/vault/vault_setup.sh
+++ b/lib/vault/vault_setup.sh
@@ -172,7 +172,7 @@ crontab -l | grep -q vault_setup
 if [[ $? -ne 0 ]]; then
   echo "creating crontab"
   crontab -l > temp_crontab
-  echo "0 0 1 * * cd $PWD; bash vault_setup.sh" >> temp_crontab
+  echo "0 0 1 * * cd $PWD; bash lib/vault/vault_setup.sh" >> temp_crontab
   crontab temp_crontab
   rm temp_crontab
 fi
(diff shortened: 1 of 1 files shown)


lib/vault/vault_setup.sh
```diff
@@ -167,3 +167,9 @@ if [ -f tmp/vault/service_stores.txt ]; then
         bash create_service_store.sh $service
     done <tmp/vault/service_stores.txt
 fi
+
+crontab -l | grep -q vault_setup
+if [[ $? -ne 0 ]]; then
+  echo "creating crontab"
+  echo "0 0 1 * * cd $PWD; bash vault_setup.sh" | crontab -
```
OrdiNeu: This will overwrite everything else in the crontab; looking online for solutions they all recommend -l but that deletes all comments

daisieh: Fixed in new commit: I saved out the old crontab and appended to that one.

OrdiNeu (approved):
Seems good for a stopgap measure -- we should think about how we might want to phase this out though (I'm not sure I've ever heard of a program automatically editing your crontab like this)

OrdiNeu (changes requested):
Wait hangon, was testing and realized I can't seem to find the vault_setup.sh script -- is it missing?

daisieh:
It is still in lib/vault/vault_setup.sh

OrdiNeu (approved):

PR #1078, opened 2025-06-02
DIG-2032: Add some debugging tips for Opa

daisieh:
Add a bit more info about debugging Opa

### DIFF
--- a/docs/src/content/docs/deployment/stack-troubleshooting.md
+++ b/docs/src/content/docs/deployment/stack-troubleshooting.md
@@ -122,3 +122,18 @@ Then recompose both tyk and federation with:
 make recompose-tyk
 make recompose-federation
 ```
+
+## Opa returns 401 Unauthorized
+
+Sometimes 401 Unauthorized errors are caused by Opa not being able to find the data it needs to validate the bearer token. These will have a message of "request rejected by administrative policy" and will not leave any decisions in the log.
+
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/deployment/stack-troubleshooting.md
```diff
@@ -122,3 +122,18 @@ Then recompose both tyk and federation with:
...
 make recompose-federation
 ```
+
+## Opa returns 401 Unauthorized
+
+Sometimes 401 Unauthorized errors are caused by Opa not being able to find the data it needs to validate the bearer token. These will have a message of "request rejected by administrative policy" and will not leave any decisions in the log.
+
+These are caused by Opa's master system.authz policy rejecting access to any downstream policies, including all of CanDIG's permission policies. Usually, this is happening because Opa can't access its Vault service store or the IDP. Try re-running `make compose-opa` to reconnect the Opa containers to Vault.
+
+If you are still having trouble diagnosing the problem, you can temporarily set Opa's system.authz authorization to allow all requests by default:
```
OrdiNeu, line 132: Recommend spelling out the file for the user, just in case:
```suggestion
If you are still having trouble diagnosing the problem, you can temporarily set Opa's system.authz (in `permissions_engine/authz.rego`) authorization to allow all requests by default:
```

OrdiNeu (approved):

PR #1079, opened 2025-06-02
Replace 'email' with 'user_id'

mshadbolt:
Replace 'email' with 'user_id' in docs for the endpoints in ingest that changed


daisieh (commented):
LGTM

PR #1083, opened 2025-06-03
Revert PR #1070: DIG-2080: add crontab to rerun vault_setup every month

OrdiNeu:
# Description
If #1077 is merged and works, we can revert #1070. I've left the code as a comment just in case we need it again.

**NB**: The crontab will still exist in everyone's systems, and will have to be removed manually.


daisieh (approved):
LGTM

PR #1085, opened 2025-06-04
DIG-2042: allow tyk to accept federated site tokens for htsget

daisieh:
This definitely works to add the server to the tyk policy file: you can see it by adding a node to the federation and then checking:
```
curl "http://candig.docker.internal:5080/tyk/apis/61" \
     -H 'x-tyk-authorization: <tmp/tyk/secret-key>'
```
where `tmp/tyk/secret-key` is whatever the value is in that file.

I haven't tested to see if this fixes downloading from federated nodes though.


mshadbolt (approved):
Working to download files from federated nodes

PR #1086, opened 2025-06-05
Hotfix: remove dud test in integration tests

daisieh:
I explained this in Slack:

> So now if we are removing a provider from tyk when we remove it from federation (as we should have always done), I see that my integration test test_add_server is faulty
It tries to add the local server again to federation, which it can do, but then delete the second one.
before, when it was NOT removing it from tyk, we’d just end up with multiple entries of the same local provider in the tyk policy. (this was something I’d noticed but never pinned down to a root cause)
but now, when it removes the provider from tyk, it has to remove all matching instances from tyk, which removes the original local server as well
Basically, this is a bad test

### DIFF
--- a/lib/federation/initialize.py
+++ b/lib/federation/initialize.py
@@ -49,14 +49,14 @@ def main():
     if response.status_code != 200:
         print(f"POST response: {response.status_code} {response.text}")
 
-    print("Making sure our server is in federation...")
-    server = get_default_server()
-    response = requests.request("POST", url, headers=headers, json=server)
-
-    if response.status_code != 200:
-        print(f"POST response: {response.status_code} {response.text}")
     url = f"{get_env_value('FEDERATION_PUBLIC_URL')}/v1/servers"
     response = requests.request("GET", url, headers=headers)
+
+    server = get_default_server()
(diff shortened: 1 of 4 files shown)


lib/federation/initialize.py
```diff
@@ -49,14 +49,14 @@ def main():
...
-    if response.status_code != 200:
-        print(f"POST response: {response.status_code} {response.text}")
     url = f"{get_env_value('FEDERATION_PUBLIC_URL')}/v1/servers"
     response = requests.request("GET", url, headers=headers)
+
+    server = get_default_server()
+    if json.dumps(server['server'], sort_keys=True) not in json.dumps(response.json(), sort_keys=True):
+        print("Adding our server to the federation")
+        response = requests.request("POST", url, headers=headers, json=server)
+        response = requests.request("GET", url, headers=headers)
```
OrdiNeu, line 59: NB: It's possible for the local server to not be entry 0 in providers with this, if they compose once with the wrong server, take federation down, and then `make compose` with the actual local server. It's not a realistic scenario though.

OrdiNeu (approved):

lib/federation/initialize.py
daisieh: Actually, since the first provider is never removed from Tyk, it doesn't matter if the local server is first in the federation list.

PR #1088, opened 2025-06-06
Add auth prefix to Katsu's docker-compose because it was missing

OrdiNeu:
Was missing from Katsu's docker-compose for some reason. Fixes a regression in #1063


daisieh (approved):
LGTM

PR #1102, opened 2025-06-17
add limit to kc

SonQBChau:
## What's new
- Set memory limit for Keycloak to reduce system resource usage
- Added restart policy to auto-restart Keycloak if the container stops

### How to test
- Tested on local and dev server


mshadbolt (approved):
worked for me in terms of mem limit. On docker desktop I can see the container is limited to 2gb memory. The memory usage on startup peaked at 1.03GB and stayed there most of the time then dropped to around 650MB after build completed. Went up to about 800MB when running integration tests.

PR #1106, opened 2025-06-19
Update the version of the workflow file for the stable documentation

OrdiNeu:
Error in the this docs deployment: https://github.com/CanDIG/CanDIGv2/actions/runs/15742622538/job/44371547909
makes me think the versions were mismatched: this is an update to it.


mshadbolt (approved):
lgtm

PR #1107, opened 2025-06-19
Bugfix: Astro update fix

OrdiNeu:
Double-checked that this one is working by running astro locally. Unsure what happened, given that the last PR I did run npm update... this time I just ran it a second time and it started to work locally?


mshadbolt (approved):
works locally for me

PR #1112, opened 2025-07-02
tiny fix: remove spaces in var assignment

daisieh:
A few variables have spaces in their assignments that might mess up `source` etc.


OrdiNeu (approved):
Looks good, didn't test but shouldn't alter behaviour

PR #1114, opened 2025-07-02
Include rnaget in example-production.env CANDIG_MODULES.

DavidBrownlee:
rnaget is required to run integration tests successfully.
rnaget is already included as a CANDIG_MODULE in both UHN and BCGSC .envs.


DavidBrownlee:
Please pause this PR.  Something else is going wacky now...

PR #1115, opened 2025-07-03
OMOP database can be started as part of CanDIG deployment

SonQBChau:
## What's new
- Added new candig-api module
- Included minimal Docker Compose setup to support make compose
- Added setup script to create OMOP DB on an existing postgres-db

### How to test
- Run candig stack
- Uncomment candig-api in .env
- make compose-candig-api

ticket DHDPS-337


mshadbolt:
Can you actually make this a separate repo instead of embedded into CanDIGv2, in case it is useful: https://github.com/CanDIG/repository-template

SonQBChau:
> Can you actually make this a separate repo instead of embedded into CanDIGv2, in case it is useful: https://github.com/CanDIG/repository-template

true, but the docker compose file and setup script are usually kept outside the sub repo

mshadbolt:
oh yeah you are right. I will review this then we can make a repo to start working on the openapi yaml

mshadbolt (approved):
LGTM- tested locally and can confrim the omop db gets created in the psql container with the omop tables

PR #1117, opened 2025-07-03
Optionally disable ROPC grants for user tokens

OrdiNeu:
# Description
[Jira ticket about ROPC removal](https://tfri.atlassian.net/browse/DHDPS-340)

Per [guidelines on OIDC](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-2.4), the usage of the Resource Owner Password Credentials (ROPC) grant is **heavily discouraged**. We've been using it for convenience and will likely still support it because it makes DevOps much easier, but with the DHDP Keycloak server not allowing ROPC we need to have an option to disable it.

The ROPC grant is what we've been using to grab access/refresh tokens in every instance except when the user logs into the portal -- ROPC allows us to sign in as users without needing to use a frontend at all. This PR allows you to remove this ability by specifying the following environment variables:
```
# Alternative to the ROPC chain -- use only if you want to disable ROPC
ENABLE_ROPC=true
AUTH_ACCEPT_PORT=8989
AUTH_ACCEPT_URL=http://${CANDIG_DOMAIN}:${AUTH_ACCEPT_PORT}
```

# To test
Please set `ENABLE_ROPC=false` in your `.env` and rebuild everything & test. **Please note that you will be prompted at several points to sign in to the given URL with the given credentials**. This may not work if you have another program listening at port 8989.

# Other notes
We don't yet have a way of doing auth_code grants *without* any browser intervention at all, which we might need for automated deployment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -539,6 +539,9 @@ print-%:
 test-integration:
 	mkdir -p tmp/test
 	python ./settings.py
+ifeq ($(ENABLE_ROPC),false)
+	source ./env.sh; python pytest-tokens.py
+endif
 ifeq ($(KEEP_TEST_DATA),true)
 	source ./env.sh; pytest -v --color=yes ./etc/tests/integration -k 'not test_clean_up' $(ARGS) --report-log=./tmp/test/test-integration_$(shell date +"%Y-%m-%d_%Hh%Mm%Ss").jsonl
 else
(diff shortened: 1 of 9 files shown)


SonQBChau (approved):
Tested and confirmed it works locally. I agree the new test integration flow is a bit cumbersome, but we don't have a better option right now

Nice work!

PR #1119, opened 2025-07-07
Fixup Keycloak crashing out if the username had a capital letter

OrdiNeu:
# Description
Keycloak crashes out if the values of any of the following variables are uppercase:
- DEFAULT_SITE_ADMIN_USER
- TEST_USER_1
- TEST_USER_2

This seems like a weird, subtle error (especially because there's no error message in any log nor in the Keycloak setup itself). I've just forced the usernames to be lowercase in this setup function -- it seems that it's able to create the users but crashes out when setting their password for some reason.


daisieh (approved):
LGTM

PR #1124, opened 2025-07-16
Data portal branch was out of date for some reason

OrdiNeu:
As noted by @dbujold the version of the data portal in stable was off for some reason -- this should fix that.


SonQBChau (approved):
approved

PR #1128, opened 2025-07-26
DHDPS-401: create external service registration

daisieh:
A service like Bitnobi may have to register itself with CanDIG so that we can be sure that calls made from their server are legitimately from the known service. You can now register an external service through the federation api and then use the authx method `verify_service_token(service, token)` to see if a call is coming from that service.

There is a test available: `pytest etc/tests/federation/test_federation.py -k test_create_client`


etc/tests/federation/test_federation.py
```diff
@@ -442,3 +500,45 @@ def test_remove_servers():
...
+    client_secret = create_keycloak_client(client_id)
+
+    username = ENV["CANDIG_NOT_ADMIN_USER"]
+    password = ENV["CANDIG_NOT_ADMIN_PASSWORD"]
+
+    user_token = get_access_token(
+    client_id=client_id,
+    client_secret=client_secret,
+    username=username,
+    password=password)
```
OrdiNeu:
```suggestion
    user_token = get_access_token(
        client_id=client_id,
        client_secret=client_secret,
        username=username,
        password=password
    )
```

etc/tests/federation/test_federation.py
```diff
@@ -127,24 +129,88 @@ def test_all_service_info():
...
+        client_id (str): The client_id for the new client
+
+    Returns:
+        client_secret
+    """
+    # Step 1: Get Keycloak admin token
+    with open(f"{REPO_DIR}/tmp/keycloak/admin-password") as f:
+        admin_pass = f.read()
+
+    # Step 2: Create user
```
OrdiNeu:
```suggestion
    # Step 2: Create client
```

OrdiNeu (commented):
Still need to test but a few starting thoughts:

etc/tests/federation/test_federation.py
daisieh: removed the whole old comment

OrdiNeu:
```
(candig) 15:14 fnguyen@fnguyen-worktop:~/Documents/CanDIGv2$ curl candig.docker.internal:5080/federation/v1/external-service -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" -d "{\"service\": \"test_client\", \"authentication\": { \"issuer\": \"http://candig.docker.internal:8080/auth/realms/candig\", \"token\": \"$CLIENT_TOKEN\" } }"
{
  "message": "Service user service-account-test_client could not be authorized: {\n    \"error\": \"Key not authorised\"\n}"
}
```

Using a valid site_admin token and a client. Does the issuer need to be added to Tyk first? Can we roll that into this function?

daisieh:
I reproduced this error and it is because this PR doesn't have the authx fix from https://github.com/CanDIG/candigv2-authx/pull/50.

daisieh:
Okay I think this is ready to test again

OrdiNeu (approved):
Reviewed in the federation PR

PR #1131, opened 2025-07-30
Prepend the site name to the email host instead

OrdiNeu:
Instead of `local_user1@test.ca` it's `user1@local_test.ca`


daisieh (approved):
Keycloak was failing to create users with the underscore in the email domain, so I switched it to a hyphen. It seems happy now.

PR #1139, opened 2025-08-19
Push candig-data-portal version to stable + add profile button fix

OrdiNeu:
This is a push to the dhdp-v6 branch that includes the Data Portal's dhdp-v6 rebase off of stable, as well as the recent profile button fix.


CourtneyGosselin (approved):
LGTM

PR #1141, opened 2025-08-26
DHDPS-414: splitting drs from htsget

daisieh:
Drs is now a standalone service, separate from htsget. To test: build the stack from the develop branch and run test-integration to get some data into the system. Then switch to this branch, making sure that your .env reflects the new example.env changes, and then run `make rebuild-keep-data` to keep the data intact. Then re-run integration tests.


OrdiNeu (approved):
LGTM :heavy_check_mark:

PR #1149, opened 2025-09-18
Doc mod for ingest-genomic.mdx

DavidBrownlee:
Additions only.


DavidBrownlee:
Really?  A .mdx documentation change breaks the build?

CourtneyGosselin:
@DavidBrownlee The Playwright test failure isn’t caused by the documentation change. It’s due to updated usernames/emails for test users. Since Playwright hasn’t been updated, the environment is still expecting the old users.

We’re not actively maintaining Playwright right now and will disable the playwright tests shortly.

DavidBrownlee:
Thank you, Courtney, for the explanation (and the next PR).

mshadbolt:
Hi @DavidBrownlee , sorry it took me so long to take a look at this and thanks for adding to our docs. I added a couple of other ways that you can monitor the indexing queue above yours in the same section. Just checking whether you knew you could also monitor the queue in this way. If you didn't know do you think you would do it by looking at the logs in future or still prefer the psql way? Not sure which one you find easier.

mshadbolt:
I consulted with the team and we don't really want to encourage people querying the psql database directly so we will stick to the two methods I added and remove the psql queries

mshadbolt (approved):
changed the docs but included more user friendly ways of checking indexing status

PR #1173, opened 2025-11-05
DIG-2113 and DIG2088: Explore section and Program Registration warning in docs website

mshadbolt:
- added warning and process for adding information to a program registration 
- added and explore section and included newly written variant search documentation


mshadbolt:
Very hastily put together explore section so would appreciate feedback on if it makes sense or any other information we should include.

daisieh (approved):
LGTM!

PR #1186, opened 2025-11-19
v7.0.0 Stable rebase

daisieh:
Merging changes back from stable to develop. Mostly env and doc updates and a few tiny stable hotfixes.


site_admin_token.py
```diff
@@ -28,15 +28,25 @@ def get_site_admin_token(username=None, password=None, refresh_token=None):
...
+                )
+        else:
+            credentials = authx.auth.get_oauth_response(
+                keycloak_url=ENV["KEYCLOAK_PUBLIC_URL"],
+                client_id=ENV["CANDIG_CLIENT_ID"],
+                client_secret=ENV["CANDIG_CLIENT_SECRET"],
+                username=username,
+                password=password,
+                refresh_token=refresh_token
+                )
```
OrdiNeu, line 49: Are we certain we want this? The old way is cleaner imo

OrdiNeu (approved):
Code looks good

docs/package.json
```diff
@@ -16,15 +16,16 @@
     "@fortawesome/fontawesome-svg-core": "^6.6.0",
     "@iconify/tailwind": "^1.1.3",
     "@lorenzo_lewis/starlight-utils": "^0.3.2",
-    "astro": "^5.15.6",
+    "astro": "^5.14.4",
```
mshadbolt:
```suggestion
    "astro": "^5.15.6",
```

docs/src/content/docs/deployment/local.mdx
```diff
@@ -283,7 +283,7 @@ make init-conda
 
 ### 3.b: Use an existing Conda installation
 
-If you want to use an existing conda installation on your local at the bottom of the [.env](../etc/env/example.env#L310), set `CONDA_INSTALL` to your existing conda installation path
+If you want to use an existing conda installation on your local at the bottom of the [.env](https://github.com/CanDIG/CanDIGv2/blob/3dcfd2b6a47135d9b632db4c4327cb75bab0768a/etc/env/example.env#L342), set `CONDA_INSTALL` to your existing conda installation path
```
mshadbolt:
```suggestion
If you want to use an existing conda installation on your local at the bottom of the [.env](https://github.com/CanDIG/CanDIGv2/blob/stable/etc/env/example.env), set `CONDA_INSTALL` to your existing conda installation path
```

etc/env/example-production.env
```diff
@@ -2,8 +2,8 @@
 # - - -
 
 # site options
-CANDIG_MODULES=logging keycloak vault redis postgres htsget katsu query tyk opa federation candig-ingest candig-data-portal
-    #drs-server wes-server monitoring
+CANDIG_MODULES=logging keycloak vault redis postgres drs htsget katsu rnaget query tyk opa federation candig-ingest candig-data-portal
+    #candig-api drs-server wes-server monitoring
```
mshadbolt:
```suggestion

```

mshadbolt (approved):
Looked over the docs and made a few suggested updates
Approved pending committing those

site_admin_token.py
daisieh: I think you're right, but I think it'd be best to make that reversion a separate PR for develop, since this version is what's in stable.

PR #1197, opened 2025-12-04
Hotfix: build ingest before federation

daisieh:
Federation needs to set up using CanDIG-authorized users, and that happens during ingest setup.


mshadbolt (approved):
Also added to production deployment page on docs, feel free to add any other notes there

Otherwise builds and passes integration tests

PR #1206, opened 2025-12-10
DRS needs access to aws secrets

daisieh:
DRS needs access to the AWS credentials stored in ingest's store in order to download files.

Already deployed on UHN prod and it works.


mshadbolt (approved):
rebuilt locally and all integration tests passed. Also tested ingest on Prod

PR #1212, opened 2025-12-22
DIG-2192: move versions out of .env file

daisieh:
I moved the versions into their own versions.env file. 

I also picked up a couple of fixes:
* build ingest before federation (this is fixed in the blocked stable-candidate but might as well go into develop)
* add a default value in some dockerfiles to prevent some meaningless errors.

To test: remove all of the versions from your .env file and then clean-all/build-all. Run integration tests.


mshadbolt (approved):
builds and tests pass

PR #1208, opened 2025-12-10
v7.1.0: Genomic search optimizations

daisieh:
* optimizations for genomic searches in query
* connected fixes for drs and htsget
* single patient genomic info (in query, not portal)

PR #1220, opened 2025-12-24
v7.1.0: Genomic search optimizations

daisieh:
- optimizations for genomic searches in query
- connected fixes for drs and htsget
- single patient genomic info (in query, not portal)

Should be the same as #1208


OrdiNeu (approved):
LGTM :heavy_check_mark:

CourtneyGosselin (approved):
LGTM

PR #1225, opened 2026-01-05
Merge v7.1.0 back to develop

daisieh:


daisieh:
Closed because I want to make this from stable-candidate instead

PR #1226, opened 2026-01-08
Tyk timeout tweaks

daisieh:
I increased the default timeout to 60 seconds to give query a bit of time to wake up. I also commented out several extraneous log messages.

### DIFF
--- a/lib/tyk/docker-compose.yml
+++ b/lib/tyk/docker-compose.yml
@@ -18,6 +18,7 @@ services:
       - "${CANDIG_DOMAIN}:${LOCAL_IP_ADDR}"
     environment:
       - DEBUG_MODE=${CANDIG_DEBUG_MODE}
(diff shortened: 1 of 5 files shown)


lib/tyk/docker-compose.yml
```diff
@@ -18,6 +18,7 @@ services:
       - "${CANDIG_DOMAIN}:${LOCAL_IP_ADDR}"
     environment:
       - DEBUG_MODE=${CANDIG_DEBUG_MODE}
+      - TYK_GW_PROXYDEFAULTTIMEOUT=60
```
OrdiNeu, line 21: Should this be a `.env` variable?

OrdiNeu (approved):

lib/tyk/docker-compose.yml
daisieh: I don't think so...I think we'd want sysadmins to not really feel like they were supposed to be adjusting it.

PR #1240, opened 2026-01-21
DIG-2080: add analysis date

daisieh:
Incorporates changes in ingest and htsget. When an analysis contains a vcf file, the htsget indexer will look for a date in the vcf headers and add it to the variantfile and then the AnalysisDrsObject. If the ingest json contains explicitly-specified analysis dates, that one will override any found in the headers.

Integration tests have been updated to check for this.


docs/src/content/docs/ingest/prepare-genomic.mdx
```diff
@@ -231,6 +232,7 @@ The file should contain an array of dictionaries, where each item represents a s
 - If an S3 bucket access method is provided, assuming you have properly added the S3 credentials to vault [(see `Configuring s3 authorization`above)](#a-configure-connection-to-genomic-data-sources), the service will scan the S3 bucket to ensure the relevant files are present.
 - There is no validation that the genomic files exist locally or are mounted to htsget. If the local (`file:///`) method is used it is important to check all files are present before proceeding with ingest.
 - `analysis_sample_id` is the name that is used inside the analysis file that is used to refer to the sample/experiment. Some common ones we see are Normal, Tumour,or may be the same identifier used for the sample. If you are not sure what this is you may need to look into the bam/vcf to see what is used.
+- `analysis_date` is the date the institute received the file from the sequencing centre, if one is not provided the CanDIG software will look for a date inside the analysis file. If one is still not found the ingest is rejected
```
mshadbolt: I added this line, is it true?

daisieh: I think this is correct

mshadbolt (approved):
tests passing and date behaviour seems to be working

OrdiNeu (approved):
LGTM ✔️

PR #1242, opened 2026-01-28
Clean up modules

daisieh:
Inspired by Marion's cleanup of CanDIGv3, a similar cleanup of v2.


mshadbolt (approved):
builds and tests pass

PR #1244, opened 2026-01-30
DIG-2240: Fix the fake error that happens during candig-ingest setup

daisieh:
If you run `make recompose-candig-ingest`, you should not see that 500 error anymore.


mshadbolt (approved):
worked for me on clean rebuild

PR #1252, opened 2026-02-14
DIG-2209: update experiment to biosample

daisieh:
We are changing the /experiments endpoint because it was confusing: it does not return something that is anything like the ENA model's [experiment](https://ena-docs.readthedocs.io/en/latest/submit/general-guide/metadata.html). We are refactoring that endpoint to something that corresponds more closely to that model.

DRS is updated to return the new endpoint; htsget is also updated to use that new endpoint and return biosamples instead of experiments.

Query and ingest are the main consumers of this endpoint, so they have to be updated to process data with the new format. Ingest only requires an update to grab the new format when calculating summary stats; any existing programs do not need to be recalculated.

Integration tests have been updated accordingly.

### DIFF
--- a/etc/tests/integration/test_integration.py
+++ b/etc/tests/integration/test_integration.py
@@ -708,25 +708,24 @@ def test_htsget_access_data(user, password, obj, access):
     assert (response.status_code == 200) == access
 
 
-def test_experiment_metadata():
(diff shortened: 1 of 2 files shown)


OrdiNeu (approved):
LGTM ✔️

PR #1283, opened 2026-03-26
DIG-2281: Ingest status needs summary statistics

daisieh:
Updates the module to https://github.com/CanDIG/candigv2-ingest/pull/199 and updates the tests to match.

### DIFF
--- a/etc/tests/integration/test_integration.py
+++ b/etc/tests/integration/test_integration.py
@@ -616,9 +616,12 @@ def test_ingest_not_admin_htsget():
     # when the user has program_curator role, they should be allowed
     print(response.text)
     assert response.status_code == 201
-    for program in response.json():
-        results = response.json()[program]
-        print(json.dumps(results["results"], indent=2))
+    resp = response.json()
+    resp.pop("last_updated")
+    resp.pop("complete")
+    for program in resp:
+        results = resp[program]
+        print(json.dumps(results, indent=2))
         for res in results["results"]:
             assert "error processing" not in res
     # clean up before the next test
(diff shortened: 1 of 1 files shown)


etc/tests/integration/test_integration.py
```diff
@@ -616,9 +616,12 @@ def test_ingest_not_admin_htsget():
     # when the user has program_curator role, they should be allowed
     print(response.text)
     assert response.status_code == 201
-    for program in response.json():
-        results = response.json()[program]
-        print(json.dumps(results["results"], indent=2))
+    resp = response.json()
+    resp.pop("last_updated")
+    resp.pop("complete")
```
kcranston: would it not make sense to also assert that complete=True, which would then cover the actual change in the ingest code?

etc/tests/integration/test_integration.py
```diff
@@ -653,9 +656,12 @@ def test_ingest_admin_htsget():
         response = requests.get(f"{ENV['CANDIG_URL']}/ingest/status/{queue_id}", headers=headers)
     # when the user has admin access, they should be allowed
     assert response.status_code == 201
-    for program in response.json():
-        results = response.json()[program]
-        print(json.dumps(results["results"], indent=2))
+    resp = response.json()
+    resp.pop("last_updated")
+    resp.pop("complete")
```
kcranston: also here, test complete=True?

kcranston (changes requested):

etc/tests/integration/test_integration.py
daisieh: sure, I can do that

kcranston (approved):

PR #1295, opened 2026-04-16
DIG-2315: add information about Vault getting too big

daisieh:
Add description of how to fix the audit file.


mshadbolt (approved):
makes sense to me

mshadbolt:
We should replicate this as merge to the stable branch too, since that is where the docs website is built from

PR #1305, opened 2026-04-17
DIG-2316: Log relevant messages from vault instead of saving vault audit log

daisieh:
Only a few transactions that are logged in the vault audit log are actually relevant to CanDIG; we added them as logged messages in authx v2.9.0 so now we can disable direct Vault audit logs.

I looked through the github test logs and it looks like all of the tokens are correctly hidden and the logging is working correctly. NB: there are logging messages like 
```
2026-04-17T18:52:29+00:00	candigv2_federation_1	{"source":"stdout","log":"level: DEBUG, file: urllib3.connectionpool, log: http://vault:8200 \"GET /v1/federation/token/9556141c-3a8e-11f1-ada1-3614914dfab3 HTTP/1.1\" 200 212"}
```
which do reveal the secret value, but those have always been there when Vault's docker-compose specifies DEBUG_MODE as true.


OrdiNeu (approved):
LGTM ✔️

PR #1306, opened 2026-04-17
update docs about vault audit log

daisieh:
Actually this belongs in stable, not develop, since this is where docs are built from.


mshadbolt (approved):
LGTM

PR #1314, opened 2026-04-24
DIG-2329 Reorganize env

SonQBChau:
## What's new
- Prod vars move at the top, make it easier to edit
- Same change to dev to make it consistent
- Rename KEYCLOAK_PRODUCTION_MODE since only KC using it

### How to test
- Copy the .env and rebuild, should work


etc/env/example-production.env
```diff
@@ -1,20 +1,58 @@
...
+TYK_ANALYTICS_FROM_EMAIL=YOUR-ADMIN-EMAIL@email.ca
+TYK_USE_SSL=true
+
+# Data portal
+CANDIG_DATA_PORTAL_PORT=4173
+CANDIG_DATA_PORTAL_URL=https://${CANDIG_DOMAIN}:${CANDIG_DATA_PORTAL_PORT}/data-portal
+CANDIG_DATA_PORTAL_SUPPORT_EMAIL=YOUR_SITE_ADMIN_SUPPORT_EMAIL
+MOHCCN_SUPPORT_EMAIL=mohdatarequests@tfri.ca
+
+DIR_PERMISSIONS=775
```
OrdiNeu: DIR_PERMISSIONS probably won't be edited super often -- it's only relevant to us sharing the tmp folder on prod.

etc/env/example-production.env
```diff
@@ -1,20 +1,58 @@
...
+KEYCLOAK_HTTPS_PORT=8443
+KEYCLOAK_PUBLIC_PROTO=https
+KEYCLOAK_PRIVATE_PROTO=http
+KEYCLOAK_PUBLIC_URL=${KEYCLOAK_PUBLIC_PROTO}://${CANDIG_AUTH_DOMAIN}
+KEYCLOAK_PRIVATE_URL=${KEYCLOAK_PRIVATE_PROTO}://keycloak:${KEYCLOAK_PORT}
+KEYCLOAK_PROXY_HEADERS=xforwarded
+
+# Tyk gateway
+TYK_LOGIN_TARGET_URL=https://${CANDIG_DOMAIN}
+TYK_ANALYTICS_FROM_EMAIL=YOUR-ADMIN-EMAIL@email.ca
```
OrdiNeu, line 43: Does anyone know how to even see this email from an instance? I see it being used in some Tyk config but I've never actually seen it referred to anywhere in any Tyk API. Might be something we replace with like, the CANDIG_DATA_PORTAL_SUPPORT_EMAIL (assumably it's the same admin account)

etc/env/example.env
OrdiNeu: Might want to move these up to the deployment config as well -- they don't change super often but CANDIG_USER_KEY might matter to some deployments (iirc the DHDP had to swap the DEFAULT_SITE_ADMIN_USER and TEST_USER_1/2)

etc/env/example.env
```diff
@@ -118,22 +147,12 @@ KATSU_THREADS=4
 
 # keycloak service
 KEYCLOAK_REALM=candig
-KEYCLOAK_CLIENT_ID=local_candig
 KEYCLOAK_LOGIN_REDIRECT_PATH=/auth/login
```
OrdiNeu, line 147: Unsure about this one and AUTH_PREFIX -- on one hand they shouldn't change so long as the user is building _our_ Keycloak, but if it's more similar to the DHDP they'll have to switch these two.

etc/env/example-production.env
SonQBChau: it is on the deployment docs, so I guess I put it on the top if people are looking for

etc/env/example-production.env
SonQBChau: I also think we don't use TYK_ANALYTICS, but if I remove it, it would require an update to Tyk and out of scope. @mshadbolt ?

etc/env/example.env
SonQBChau: These are not in the deployment docs. I tried to move only what relevant to the list here https://candig.github.io/CanDIGv2/deployment/production/#env-settings and keep both dev and prod env the same since people run the diff on them. If this is a config vars we should add it to the docs and I will move it top. @mshadbolt ?

etc/env/example.env
SonQBChau: Again, this is not in the docs, so I'm also unsure if we want to make it for exceptional case? @mshadbolt

SonQBChau:
@OrdiNeu thank for the feedback! I asked @mshadbolt to clarify some comments and will update it accordingly.
I also want to mention that most of the changes doesn't apply to dev, but I move them to the top as well since I want to keep the 2 identical for later edit

daisieh (commented):
Seems to me that there is some confusion about whether or not everything that is in the docs is actually something we need people to know about because their values sometimes change, or if we just have them in there because they are visible and might confuse people.

etc/env/example-production.env
```diff
@@ -33,10 +71,6 @@ TEST_USER_2=user2@${CANDIG_SITE_LOCATION}-test.ca
 # Set KEEP_TEST_DATA to true to have the integration tests keep all test data
 KEEP_TEST_DATA=false
 
-# set this to your local IP address (i.e. 192.168.x.x on most networks) if `make init-authx` cannot automatically determine your IP
-LOCAL_IP_ADDR=
- # may need to set this value
-
 # miniconda venv
 # options are [linux, darwin, arm64mac]
 VENV_OS=linux
```
mshadbolt: Can this also move up, probably just below local_ip_addr, it is the one I change all the time

etc/env/example-production.env
```diff
@@ -1,20 +1,58 @@
...
 CANDIG_AUTH_DOMAIN=YOUR_PRODUCTION_AUTH_DOMAIN
 CANDIG_SITE_LOCATION=YOUR_SITE_IDENTIFIER
- # BCGSC, UHN, C3G, etc.
 CANDIG_DEBUG_MODE=0
-CANDIG_PRODUCTION_MODE=1
+LOCAL_IP_ADDR=
+
+# Federation
+FEDERATION_PORT=4232
+FEDERATION_SERVICE_URL=http://${CANDIG_INTERNAL_DOMAIN}:${FEDERATION_PORT}
```
mshadbolt, line 26: would they need to change these?

etc/env/example.env
```diff
@@ -1,21 +1,57 @@
...
+# Federation
+FEDERATION_PORT=4232
+FEDERATION_SERVICE_URL=http://${CANDIG_INTERNAL_DOMAIN}:${FEDERATION_PORT}
+FEDERATION_SELF_SERVER_ID=internal-1
+FEDERATION_SELF_SERVER="{'id': '${FEDERATION_SELF_SERVER_ID}', 'url': '${FEDERATION_SERVICE_URL}','location': {'name': '${CANDIG_SITE_LOCATION}','province': 'ON','province-code': 'ca-on'}}"
+
+# Keycloak
+KEYCLOAK_PRODUCTION_MODE=0
+KEYCLOAK_CLIENT_ID=local_candig
+KEYCLOAK_PORT=8080
```
mshadbolt, line 34: would they ever change the port?

etc/env/example.env
```diff
@@ -1,21 +1,57 @@
...
+KEYCLOAK_HTTPS_PORT=8443
+KEYCLOAK_PUBLIC_PROTO=http
+KEYCLOAK_PRIVATE_PROTO=http
+KEYCLOAK_PUBLIC_URL=${KEYCLOAK_PUBLIC_PROTO}://${CANDIG_AUTH_DOMAIN}:${KEYCLOAK_PORT}
+KEYCLOAK_PRIVATE_URL=${KEYCLOAK_PRIVATE_PROTO}://${CANDIG_AUTH_DOMAIN}:${KEYCLOAK_PORT}
```
mshadbolt, line 39: I don't think these would ever get changed since they are just chaining together the other vars

etc/env/example.env
```diff
@@ -1,21 +1,57 @@
...
+KEYCLOAK_PROXY_HEADERS=none
+
+# Tyk gateway
+TYK_LOGIN_TARGET_URL=http://${CANDIG_DOMAIN}:${TYK_SERVICE_PUBLIC_PORT}
```
mshadbolt, line 43: would this ever get changed?

etc/env/example.env
```diff
@@ -1,21 +1,57 @@
...
+TYK_ANALYTICS_FROM_EMAIL=admin@distributedgenomics.ca
+TYK_USE_SSL=false
+
+# Data portal
+CANDIG_DATA_PORTAL_PORT=4173
```
mshadbolt, line 48: would this get changed?

etc/env/example.env
```diff
@@ -34,9 +70,6 @@ TEST_USER_2=user2@${CANDIG_SITE_LOCATION}-test.ca
 # Set KEEP_TEST_DATA to true to have the integration tests keep all test data
 KEEP_TEST_DATA=false
```
mshadbolt: move this near the top since it gets changed often

etc/env/example.env
```diff
@@ -1,21 +1,57 @@
...
+CANDIG_DATA_PORTAL_URL=https://${CANDIG_DOMAIN}:${CANDIG_DATA_PORTAL_PORT}/data-portal
+CANDIG_DATA_PORTAL_SUPPORT_EMAIL=site_admin@test.ca
+MOHCCN_SUPPORT_EMAIL=mohdatarequests@tfri.ca
+
+DIR_PERMISSIONS=775
+
+# =============================================================================
+# END DEVELOPMENT CONFIGURATION
+# =============================================================================
+
```
mshadbolt, line 56: Maybe put a note here something like 
```suggestion
Variables in this section are unlikely to need to be changed. Please consult with a CanDIG developer before changing if you aren't familiar with what the variable does.
```

etc/env/example.env
mshadbolt: I think CANDIG_USER_KEY should probably move up since we don't necessarily know what idp the user is using. I am not sure about the site admin user

etc/env/example.env
```diff
@@ -34,9 +70,6 @@ TEST_USER_2=user2@${CANDIG_SITE_LOCATION}-test.ca
 # Set KEEP_TEST_DATA to true to have the integration tests keep all test data
 KEEP_TEST_DATA=false
 
-# set this to your local IP address (i.e. 192.168.x.x on most networks) if `make init-authx` cannot automatically determine your IP
-LOCAL_IP_ADDR=
-
 # miniforge venv
 # options are [linux, darwin, arm64mac]
 VENV_OS=linux
```
mshadbolt: move near top

etc/env/example.env
mshadbolt: I don't think we should consider DHDP when we are doing this since we are unlikely to encounter that kind of deployment for CanDIGv2

mshadbolt (changes requested):
A few changes inline comments. Not sure for some of the ports and things if they would ever change.

etc/env/example-production.env
SonQBChau: No, but they need to be up top because FEDERATION_SELF_SERVER refers them

etc/env/example.env
SonQBChau: Same reason, they are referenced

etc/env/example.env
SonQBChau: Same reason, they are referenced

etc/env/example.env
SonQBChau: Same reason, they are referenced

etc/env/example.env
SonQBChau: Same reason, they are referenced

SonQBChau:
- move DIR_PERMISSIONS to the bottom again
- move CANDIG_USER_KEY up
- move VENV_OS up
- move KEEP_TEST_DATA up

etc/env/example-production.env
OrdiNeu: Re-pinging this because I still don't know if we want this

etc/env/example-production.env
SonQBChau: moved

etc/env/example.env
SonQBChau: moved CANDIG_USER_KEY

etc/env/example-production.env
SonQBChau: moved

etc/env/example.env
daisieh: Shouldn't this information be at the start of the section, not the end? It seems confusing this way, as if it applies to the section following, not the section before.

daisieh (approved):
Seems cleaner to me. I added a question about the comment that Marion added to the development section.

SonQBChau:
I moved the comments up as Daisie requested

mshadbolt (approved):
lgtm

PR #1315, opened 2026-04-27
DIG-2282: Set last verification timestamp as metadata in AnalysisDrsObjects

daisieh:
The new integration test should verify that file verification works as expected.


OrdiNeu (approved):
✔️ LGTM

PR #1323, opened 2026-04-29
DIG-2308: Move full paths into the modules

daisieh:


daisieh:
closing in favor of #1343

PR #1332, opened 2026-05-12
Be a bit more clear when federation has failed to add anything

OrdiNeu:
This replaces the success message when federation has not added any new servers, as it always kind of looked strange that it reports success when nothing's been federated.


SonQBChau (approved):
lgtm

PR #1350, opened 2026-05-21
Fixup federation tests' urls

OrdiNeu:
#1343 changes how federation fanout calls' paths work, so this updates them in the federation tests.

### DIFF
--- a/etc/tests/federation/test_federation.py
+++ b/etc/tests/federation/test_federation.py
@@ -358,7 +358,7 @@ def test_querying_site_query_authorized_remote_test_dataset():
     # Step 1: can we do a discovery query successfully to all sites?
     body = {
         "method": "GET",
-        "path": "discovery/programs",
+        "path": "query/discovery/programs",
         "payload": {},
         "service": "query"
     }
@@ -384,7 +384,7 @@ def test_querying_site_query_authorized_remote_test_dataset():
     programs -= {"TEST_FEDERATE"}
     body = {
         "method": "GET",
-        "path": "query",
+        "path": "query/query",
         "payload": {"exclude_programs": ",".join(programs)},
         "service": "query"
     }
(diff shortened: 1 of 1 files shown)


etc/tests/federation/test_federation.py
```diff
@@ -89,10 +89,10 @@ def test_all_service_info():
     }
     endpoints = {
         # all of these endpoints should return JSON
-        "htsget": f"ga4gh/drs/v1/service-info",
+        "htsget": f"{ENV['CANDIG_ENV']['TYK_HTSGET_API_LISTEN_PATH']}/beacon/v2/service-info",
         "katsu": f"v3/service-info",
```
daisieh, line 93: Does Katsu also need its listen path?

daisieh (approved):
quick q, otherwise looks good.

etc/tests/federation/test_federation.py
OrdiNeu: I've tried it a few times with `katsu/v3/service-info` and it always fails -- did Katsu not get affected by the filepaths thing?
```
[        
  {
    "location": {
      "name": "demo",                                                                       
      "province": "ON",        
      "province-code": "ca-on"
    },              
    "message": "handle_server_request failed on dev-1, federation = false: {'results': {}, 'service': 'katsu', 'status': 500}",
    "service": "katsu",
    "status": 500  
  },
  {                                                                                         
    "location": {                                                                                                                                                                       
      "name": "federation-1",                                                                                                                                                           
      "province": "ON",         
      "province-code": "ca-on"                                                              
    },
    "message": "handle_server_request failed on federation-1, federation = false: {'results': {}, 'service': 'katsu', 'status': 500}",
    "service": "katsu",                                                                     
    "status": 500                                                                           
  }                                                                                         
]
```
Conversely, it works fine with just `v3/service-info`

PR #1354, opened 2026-05-21
DIG-2308: update various npm packages and adds openapi snippets

mshadbolt:
Updated various astro/starlight packages
  * applied a patch from Claude so that plugins still worked
Added curl code snippets that show the full path for all the technical API documentation
Added a page that describes the openapi docs a little at `technical/api-docs`

To test locally, 
delete you node_modules folder and package-lock.json if you have done this before

```
cd docs
npm install
npm run dev
```


daisieh (approved):
LGTM

mshadbolt:
merging, noting this won't be deployed to the docs website until we make a stable, and at that point we should also update all the openapi schemas to the stable versions as well

PR #1358, opened 2026-05-30
DIG-2344: use as_self param when initializing our server

daisieh:
Added an as_self boolean parameter to the federation POST /server endpoint, so that when we add our server to the federation (as we do on initialization), we mark it as our self. When returning servers with get_registered_servers(), return self as the first server.

To test: /service-info should return our own server. If you then POST to /server with a different server and use the as_self param, you should then get that other server from the service-info endpoint. If you then POST to /server with your own server as_self again, you should then get your own server back at the service-info endpoint.


OrdiNeu (approved):
LGTM ✔️

PR #1372, opened 2026-06-05
remove docs, setup redirect

mshadbolt:
Removed docs from the repo and set up redirect to the new website

To test, go to old docs website and you should be redirected to the new one.

Will need to remove this branch from the workflow file before merging


daisieh (approved):
LGTM

PR #1388, opened 2026-06-16
DIG-2372: add compose-query to restore-vault

daisieh:
Query needs to run its setup script again to make sure it's got the correct tokens for vault etc.

`make clean-all`, `make test-integration`, `make backup-vault`.
`make clean-all`, `make build-all`, use the restore from before and `make restore-vault`.
Tests should now pass.


mshadbolt (approved):
working for me

PR #1389, opened 2026-07-03
DIG-2378: make recipes for backing up and restoring postgres scripts

daisieh:
* Created new scripts in lib/postgres to backup and restore postgres databases:
   * backup_db.sh takes the name of the database as an argument and creates a backup file in the $BACKUP_LOCATION directory
   * restore_db.sh takes the name of the database and the name of the module as arguments, looks for a file named restore.txt in the lib/module directory, and restores from the sql file path listed in that file. (this is a path instead of a file because these sql files can be HUGE).
* Created some new variables in example.env: added vars for `DRS_DB`, `HTSGET_DB`, `KATSU_DB`, and `RNAGET_DB`. `CANDIG_DB_NAMES` lists those db names, while `CANDIG_DB_MODULES` specifies which modules use postgres dbs.
   * updated settings.py to save these names in env.sh
   * replaced direct references in docker-compose files etc to these variables
* When a preflight script gets run as part of `build-%`, restore_db.sh is run with the relevant arguments.
* Added Makefile targets for `backup-all-postgres`, `restore-postgres-%`, and `restore-all-postgres`, to make it easier for admins to run these correctly.

To test: run `make backup-all-postgres` on a local instance with test data. Note that the backup files are created in tmp/backups or wherever the $BACKUP_LOCATION is specified. Save these off somewhere else. 
Clean and build-all from scratch. Then create files in lib/$*/restore.txt, each containing the full path to the relevant backup file. When you run `make restore-all-postgres`, you should notice a bunch of psql commands being run. Integration tests should work again.

### DIFF
--- a/lib/postgres/backup_db.sh
+++ b/lib/postgres/backup_db.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+python settings.py
+source env.sh
+
+db_name=$1
+
+postgres=$(docker ps -a --format "{{.Names}}" | grep postgres-db | awk '{print $1}')
+PGDATA=$(docker exec $postgres sh -c 'echo $PGDATA')
+
+echo "Making backup of ${db_name} database..."
+docker exec $postgres sh -c "pg_dump -U admin -d ${db_name} -f ${PGDATA}/${db_name}-backup.sql"
+if [[ $? -eq 0 ]]; then
+    docker cp $postgres:$PGDATA/${db_name}-backup.sql $(pwd)/$BACKUP_LOCATION/${db_name}-backup.sql
+fi
+docker exec $postgres sh -c "rm ${PGDATA}/${db_name}-backup.sql"
+
+echo
+echo "Saved $db_name backup to $BACKUP_LOCATION"
+ls -l $(pwd)/$BACKUP_LOCATION/${db_name}-backup.sql
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,7 @@ lib/minio/aws-credentials
 !/.git/
 !.gitignore
 !.gitmodules
+
+# Ignore restore/restored files
+lib/*/restore.*
+lib/*/restored.*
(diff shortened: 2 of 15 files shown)


mshadbolt:
wondering if it would also make sense to have a target to backup everything as well as separate ones for vault and postgres? Like I feel like `make backup-all` and `make restore-all` would be most convenient

lib/postgres/backup_db.sh
```diff
@@ -0,0 +1,20 @@
...
+python settings.py
+source env.sh
+
+db_name=$1
+
+postgres=$(docker ps -a --format "{{.Names}}" | grep postgres-db | awk '{print $1}')
+PGDATA=$(docker exec $postgres sh -c 'echo $PGDATA')
+
+echo "Making backup of ${db_name} database..."
+docker exec $postgres sh -c "pg_dump -U admin -d ${db_name} -f ${PGDATA}/${db_name}-backup.sql"
```
mshadbolt, line 12: wondering if we should add a timestamp to the backup filenames?

daisieh: I don't like doing that because then if the admin is not paying attention, they might just clog up their BACKUP_LOCATION with a billion backups.

yavyx:
Tested and working as expected, backed up and restored successfully with files saved externally.

Two checks I think should be added:
-  I had an old .env file with no `CANDIG_DB_NAMES` and when running `make backup-all-postgres` I get `Copy these backup files to an archival location!` even though no backup files were made. The script should check if the variables are there and inform the user if they aren't.
- I changed the file names in `lib/$*/restore.txt` to non-existent files. `make restore-all-postgres` ran with no errors and it didn't restore any data (as expected) but there should be a check that ensures the files in `restore.txt` exist and stops the script if they don't.

daisieh:
Okay I think I've fixed your two requests. Test again?

yavyx (approved):
Tested both issues and working as expected, LGTM!

PR #1390, opened 2026-07-04
DIG-2353: add warning to destructive make commands

daisieh:
For production systems (ones with CANDIG_DEBUG_MODE=0 set in their .env files), warn the user before performing any destructive actions.

To test, set your CANDIG_DEBUG_MODE to 0 in your .env file, then run any clean target, like `make clean-all`. You should be required to enter "I UNDERSTAND" before proceeding; if you don't type exactly that, it won't continue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,23 @@ CONDA = $(CONDA_INSTALL)/bin/conda
 CONDA_ENV_SETTINGS = $(CONDA_INSTALL)/etc/profile.d/conda.sh
 
 
+# this target prompts the user to confirm that they understand; the .INTERMEDIATE target designates the touched file as needing to be cleaned up at the end of make.
+# it should only need to be run on non-debug-mode
+warn:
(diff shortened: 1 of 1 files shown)


mshadbolt:
The main feedback I have is that the warning is kind of scary and doesn't always apply, i.e. if I recompose some services, it doesn't actually delete data, e.g. candig-ingest, candig-data-portal, query but it always says I will delete data which makes me scared to do it, even though I know it doesn't actually have an impact.

daisieh:
But does it hurt to have an overly-scary warning for production systems?

daisieh:
If you could suggest a less-scary warning that would apply across the compose-% targets, I'd be happy to change the wording, but I can't make a warning that only affects some of the compose-% targets.

mshadbolt:
Yeah that's true, I guess now that we have the backup targets, backing up is also much easier so doesn't hurt to have extra backups

mshadbolt (approved):
lgtm

PR #1387, opened 2026-06-11
DIG-2367: check for suspiciously small vault restore file

daisieh:
Do a clean install on your local system. Run `make backup-vault` on the fresh system to get a tiny backup file: `tmp/backups/vault.tar.gz` should be less than 50kb in size.

Copy that file to `lib/vault/restore.tar.gz` to trigger a restore-from-backup and then run `make restore-vault`. Watch the progress. When you get to "restoring vault from backup", you should get "BAD RESTORE FILE" etc. The file should be moved to `not_restored.tar.gz`.

Run integration tests to get a small amount of data into your vault and then run `make backup-vault` again. This file should now be bigger than 50kb.

Copy that file to `lib/vault/restore.tar.gz` to trigger a restore-from-backup and then run `make restore-vault`. This time, vault should be restored from backup and you should be able to use the site admin token to see data:
```
curl "http://candig.docker.internal:5080/ingest/program" \
     -H 'Authorization: Bearer <site admin token>'

[
  "PROGRAM-2",
  "PROGRAM-1",
  "TEST_2",
  "TEST_3",
  "local-SYNTH_01",
  "local-SYNTH_02",
  "local-SYNTH_03",
  "local-SYNTH_04"
]

```

PR #1391, opened 2026-07-04
DIG-2400: check for recent backups

daisieh:
When running make build-all (or any other target that calls the pre-build-check script), make sure that the user sees the size and dates of the latest backups and confirms that they are good.

Set CANDIG_DEBUG_MODE=0 in your .env file and then run `make build-all`.

If you have never backed anything up, you will get
```
Backup files located at tmp/backups:
total 0
Do these backups look correct and up to date? (y/n)
```

Go ahead and finish building and running tests; then create some backups with `make backup-vault` and `make backup-all-postgres`. When you try to run `make build-all` again, you should see the contents of that backup location before being prompted as before.

### DIFF
--- a/pre-build-check.sh
+++ b/pre-build-check.sh
@@ -84,7 +84,7 @@ fi
 if [[ $CANDIG_DEBUG_MODE == 0 ]]; then
     echo
     echo "Backup files located at ${BACKUP_LOCATION}:"
-    ls -lh $BACKUP_LOCATION
+    ls -lh $BACKUP_LOCATION | tail -n +2
     while [[ "$SILENT_MODE" != 1 ]]
     do
         read -r -p 'Do these backups look correct and up to date? (y/n) ' choice
--- a/pre-build-check.sh
+++ b/pre-build-check.sh
@@ -42,7 +42,7 @@ else
     echo "$DIFF_OUT"
     while [[ "$SILENT_MODE" != 1 ]]
     do
-        read -r -p 'Do you want to continue? (y/n)' choice
+        read -r -p 'Do you want to continue? (y/n) ' choice
         case "$choice" in
           n|N) exit 1;;
           y|Y) break;;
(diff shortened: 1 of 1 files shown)


pre-build-check.sh
```diff
@@ -80,3 +80,18 @@ else
     fi
 fi
 
+# Check 7: (ONLY FOR PRODUCTION) Are there recent backups?
+if [[ $CANDIG_DEBUG_MODE == 0 ]]; then
+    echo
+    echo "Backup files located at ${BACKUP_LOCATION}:"
```
mshadbolt, line 86: Where does the $BACKUP_LOCATION variable come from?

daisieh: .env file; but I think I put its specification in the pull request I gave to Javi: https://github.com/CanDIG/CanDIGv2/pull/1389

daisieh: Weirdly it seems to be in this commit but is not showing in this PR: https://github.com/CanDIG/CanDIGv2/commit/690eef2d70ce1b1d093b4b353f0cccb4d7ea12a0

daisieh: Actually that was merged with https://github.com/CanDIG/CanDIGv2/pull/1387.

mshadbolt:
What is the `total <number>` supposed to indicate?

For example, I have one vault backup and I get: 
```
Backup files located at tmp/backups:
total 120
-rwxrwxr-x  1 marionshadbolt  staff  61440 Jun 16 16:35 vault.tar.gz
```

Don't really get what `120` means? or what I should be looking for in terms of the size? Maybe we can print the human readable style when we do the ls command as well? i.e add `-h`

pre-build-check.sh
```diff
@@ -80,3 +80,18 @@ else
...
+    ls -l $BACKUP_LOCATION
```
mshadbolt:
```suggestion
    ls -lh $BACKUP_LOCATION
```

add human readable so the bytes are easier to read

daisieh:
I have never actually considered what that `total xxx` line means in ls! According to the man page:
```
The listing of a directory's contents is preceded by a labeled total number of blocks used in the file system by the files which are listed as the directory's contents (which may or may not
     include . and .. and other files which start with a dot, depending on other options).
```

yavyx:
Prompt works as expected! However, isn't it more important to show it before `make clean-all` since that's the destructive step where data is lost? I think this is what Courtney meant too.

daisieh:
There should be a warning when you run make clean-all: https://github.com/CanDIG/CanDIGv2/pull/1390

daisieh:
(I need to rebase this branch on the new develop, one sec)

daisieh:
forgot to rebase this one from develop

yavyx (approved):
Tested and working! Got make-clean-all warning

PR #1395, opened 2026-07-14
Bump pip version and remove irrelevant line from makefile

mshadbolt:
Pip version was not compatible with python 3.14
Line in makefile shouldn't be needed anymore since we upgraded python and pip


daisieh (approved):
Works in the github action, looks fine to me

PR #1397, opened 2026-07-16
DIG-2406: program dac auths don't propagate correctly

daisieh:
There is a fix in candig-ingest and some tests to test the fix.


mshadbolt (approved):
lgtm

PR #1400, opened 2026-07-17
DIG-2405: site roles should only be accessible by admins

daisieh:
Added fixes and tests


mshadbolt (approved):
LGTM confirmed site curator cannot add/edit/view/delete site admin, including themself

PR #1409, opened 2026-07-22
Update nightly_build.sh

SonQBChau:
## What's new
- Update nightly_build.sh to avoid conflict
- The old script using stash pop to reapply the local changes causing conflict
- New script would use the latest develop to build itself


mshadbolt (approved):
approving so we can see if it works overnight

PR #1413, opened 2026-07-23
Disable conda build and activate within nightly build

daisieh:
I updated the cronjob on dev from
```
  24 0 4 * * * cd /opt/candig/CanDIGv2; ./nightly_build.sh
```
to 
```
  24 0 4 * * * cd /opt/candig/CanDIGv2; conda activate candig; ./nightly_build.sh
```
so we don't need to re-install conda and activate inside the script.


SonQBChau (approved):
👍

PR #1416, opened 2026-07-27
Pass VITE_AGGREGATE_COUNT_THRESHOLD as portal build arg

mshadbolt:
## Problem
The data portal's censorship caption rendered "Totals do not include counts of less than **undefined**" because `VITE_AGGREGATE_COUNT_THRESHOLD` was never supplied to the portal at build time. Vite bakes `import.meta.env` in at build time, but the value was only passed as a runtime `environment` var.

## Change
Pass `AGGREGATE_COUNT_THRESHOLD` into the `candig-data-portal` build `args` as `VITE_AGGREGATE_COUNT_THRESHOLD`, matching the other `VITE_*` build args. The submodule Dockerfile declares the corresponding `ARG`/`ENV`.

## Companion PR
Pairs with candig-data-portal PR #269 (Dockerfile `ARG`/`ENV` + config fallback + the Highcharts #14 fixes). That submodule PR should merge/tag first, then the submodule pointer here be bumped.

## Testing
Rebuilt the portal locally (`make build-candig-data-portal && make compose-candig-data-portal`) with `AGGREGATE_COUNT_THRESHOLD=10`; caption reads "…less than 10…".


daisieh (approved):
lgtm

PR #1417, opened 2026-07-28
Pass VITE_CANDIG_VERSION as portal build arg (footer version)

mshadbolt:
## Problem
The data portal footer rendered "CanDIG undefined" and linked to `releases/tag/undefined` because `VITE_CANDIG_VERSION` was never supplied to the portal at build time — it was only passed as a runtime `environment` var, which Vite ignores.

## Change
Pass `CANDIG_VERSION` into the `candig-data-portal` build `args` as `VITE_CANDIG_VERSION`, matching the other `VITE_*` build args. The submodule Dockerfile declares the corresponding `ARG`/`ENV`.

## Companion PR
Pairs with candig-data-portal PR #271 (Dockerfile `ARG`/`ENV`). Same pattern as the recently merged threshold fix (CanDIGv2 #1416 / candig-data-portal #269). That submodule PR should merge/tag first, then the submodule pointer here be bumped.

## Testing
Rebuilt locally with `CANDIG_VERSION=v8.0.0`; footer reads "CanDIG v8.0.0".


daisieh (approved):
lgtm

PR #1419, opened 2026-08-12
Hotfix: remove duplicate federation values

daisieh:
These values were moved up to the top of the file (lines 28-29), so I'm removing their duplicated entries.


daisieh:
The new migrate-env script will update all the instances anyway, so this shouldn't make any difference for any sites right now, but it would be nice to not have duplicates that could get out of sync.

mshadbolt (approved):
lgtm

PR #1420, opened 2026-08-13
Hotfix: if it's a production env, don't make synth data

daisieh:
I'm wondering if this step makes people think that the ingest build is hanging...it can take awhile.


mshadbolt:
I thought they still need synth data if they want to run integration tests? Or do the production integration tests not use the synthetic data?

daisieh:
The production integration tests don't use synthetic data. @SonQBChau can you confirm?

SonQBChau:
yep, the prod test doesn't use synthetic data. 
To clarify: `make test-integration` will use it. `make test-integration-prod` doesn't

mshadbolt:
so we don't think it is important to test ingest on prod?

daisieh:
We can't really; we've never done it before because it's sketchy to contaminate the production db with test data.

mshadbolt (approved):
LGTM

